PR #28, opened 2019-03-19
Miscellaneous cleanup (pep8, test suite, comparisons)

klauer:
* Fixed pytest 4.0-related deprecations and unpinned it in requirements
* Removed trailing whitespace and other miscellaneous pep8 fixes (committed as @n-wbrown as I had nothing to do with the code)
* Removed unused test_initial (created from the cookiecutter)
* Fixed most PEP8-related errors
* String comparisons using `is` and `None` comparisons using `==` have generally been fixed

There's one string comparison which causes `test_TmcFile_fullbuild` to fail when it's corrected. I don't know enough about the logic surrounding it to figure it out just yet...

I'd recommend looking at the diff with whitespace changes ignored due to the nature these style PRs.


n-wbrown:
Thanks @klauer! I've started an issue for the test_TmcFile_fullbuild issue you've found.

PR #42, opened 2019-04-15
ENH: validate generated db file against IOC dbd using pyPDB linter

klauer:
With this PR, `pyPDB` is now a runtime dependency. 

A `.dbd` file can optionally be passed to `pytmc` to validate its contents prior to writing the output file.

(The original contents of this PR have been wiped away and refactored due to the agreement on the pyPDB dependency)


pytmc/bin/pytmc.py
```diff
@@ -0,0 +1,121 @@
...
+                        if dbd_file is not None
+                        else pytmc.fields.valid_fields)
+        bad_fields = tmc_file_obj.validate_record_fields(valid_fields)
+        for record, field in bad_fields:
+            pv_name = record.cfg_as_dict().get('pv')
+            logger.error('Record: %r has invalid field specified: %r', pv_name,
+                         field)
+            record.cfg.remove_config_field(field)
+
+        if bad_fields and not allow_invalid_fields:
```
teddyrendahl: Do we want the default behavior to fail here? The IOC will still work right? Warning might be enough, then for release we do the stricter build?

pytmc/fields.py
```diff
@@ -0,0 +1,225 @@
+# - NOTE -
```
teddyrendahl: Why have this be a Python file? Why not a JSON, just makes it more intuitive that it is a generated file?

pytmc/bin/pytmc.py
klauer: My opinion is that it's a safer default to fail, requiring the user to explicitly ignore it.

pytmc/fields.py
klauer: I have no strong opinions about this. If JSON is your preference, that can be done as well.

The larger question here is: should there be a runtime dependency on `pypdb`, obviating the need for this file altogether?

pytmc/fields.py
```diff
@@ -0,0 +1,225 @@
...
+         'F64IV', 'F64OUT', 'FCTL', 'GPIBIV', 'HOSTINFO', 'I32INP', 'I32IV',
+         'I32OUT', 'IEOS', 'IFACE', 'IFMT', 'IMAX', 'IXANY', 'IXOFF', 'IXON',
+         'LBAUD', 'MCTL', 'NAWT', 'NORD', 'NOWT', 'NRRD', 'OCTETIV', 'OEOS',
+         'OFMT', 'OMAX', 'OPTIONIV', 'PCNCT', 'PORT', 'PRTY', 'REASON', 'SBIT',
+         'SPR', 'TB0', 'TB1', 'TB2', 'TB3', 'TB4', 'TB5', 'TFIL', 'TIB0',
+         'TIB1', 'TIB2', 'TINB0', 'TINB1', 'TINB2', 'TINB3', 'TINM', 'TINP',
+         'TIOM', 'TMOD', 'TMOT', 'TMSK', 'TSIZ', 'UCMD', 'UI32INP', 'UI32IV',
+         'UI32MASK', 'UI32OUT', 'VAL']
+    ),
+    'axis': common_fields.union(
```
ZLLentz: We should be generating the `motor` fields instead of the `axis` fields since we're committed to going full motor motor

ZLLentz:
I like the ideas you have regarding stronger tie-ins of dbd files

pytmc/fields.py
klauer: Good catch - we should update the package-provided default `.dbd` to be the most complete one we have available. Any suggestions?

ZLLentz: The one that builds with the most recent motor record should be sufficient, right?

pytmc/bin/pytmc.py
teddyrendahl: Seems fair.

pytmc/fields.py
klauer: I think so. If there are any other records in even somewhat common usage, we might want to add onto it more though.

klauer:
This also doesn't touch on validating what's in the fields. While rather important, we can't catch everything. I think some limited tests on field values (e.g., is DTYP valid according to the dbd?) would be advisable down the line.

pytmc/fields.py
ZLLentz: I don't know of any other special records that we want to use with the Beckhoff PLCs

pytmc/fields.py
klauer: OK, this is now removed due to the refactor

klauer:
(Sorry, I think I'm finally done with the dumb force pushes.)

Ready for a second round of reviews.

pytmc/bin/pytmc.py
```diff
@@ -31,6 +31,20 @@ def make_db_text(tmc_file_obj):
...
+    if dbd_file is not None:
+        results = tmc_file_obj.validate_with_dbd(dbd_file)
+        for warning in results.warnings:
+            logger.warning('[%s line %s] %s', warning['file'], warning['line'],
+                           warning['message'])
+        for error in results.errors:
+            logger.error('[%s line %s] %s', error['file'], error['line'],
+                         error['message'])
+        if not results.success and not allow_errors:
+            logger.error('Linter errors - failed to create database. '
```
teddyrendahl: Think this error message might be a little misleading? We created a database it is just invalid?

pytmc/xml_collector.py
```diff
@@ -759,6 +808,32 @@ def motor_record_as_dict(self):
         """
         raise NotImplementedError
 
+    @property
```
teddyrendahl: 👍

pytmc/xml_collector.py
```diff
@@ -759,6 +808,32 @@ def motor_record_as_dict(self):
...
+        pvname : str or None
+            The record type associated with the BaseRecordPackage
+        """
+        try:
+            return self.cfg.get_config_lines('type')[0]['tag']
+        except (TypeError, KeyError):
+            pass
+
+    @property
+    def pvname(self):
```
teddyrendahl: 👍

pytmc/bin/pytmc.py
teddyrendahl: Are you sure you don't want to write the database still? It might be nice to go look at why the linter is yelling at you?

pytmc/epics.py
```diff
@@ -0,0 +1,172 @@
...
+    def __init__(self, fn):
+        '''
+        Load an expanded EPICS dbd file
+
+        Parameters
+        ----------
+        fn : str or file
+            dbd filename
+
+        Returns
```
n-wbrown: A little puzzled. Is __init__ returning something? Is there a __call__ method here somewhere? Did I miss something?

pytmc/xml_collector.py
```diff
@@ -368,6 +368,55 @@ def configure_packages(self):
...
+        results = epics.lint_db(dbd=dbd_file, db=self.render(),
+                                **linter_options)
+        if remove_invalid_fields:
+            all_invalid_fields = [
+                error['format_args']
+                for error in results.errors
+                if error['name'] == 'bad-field'
+                and len(error['format_args']) == 2
+            ]
+            print(all_invalid_fields)
```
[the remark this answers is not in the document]
klauer: Mistake - debug print

pytmc/epics.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Parameters
+    ----------
+    dbd : DbdFile or str
+        The database definition file; filename or pre-loaded DbdFile
+    db : str
+        The database filename or text
+    full : bool, optional
+        Validate as a complete database
+    warn_quoted : bool, optional
```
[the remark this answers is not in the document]
klauer: If false, it'll be silent

pytmc/epics.py
klauer: Good catch, copy/paste error

pytmc/xml_collector.py
```diff
@@ -368,6 +368,55 @@ def configure_packages(self):
...
+            invalid_fields_by_record = defaultdict(set)
+            for record_type, field_name in all_invalid_fields:
+                invalid_fields_by_record[record_type].add(field_name)
+
+            for pack in self.all_RecordPackages:
+                for field in invalid_fields_by_record.get(pack.record_type,
+                                                          []):
+                    pack.cfg.remove_config_field(field)
```
n-wbrown: I think it might be nice to have a way to track these. An incorrect record that needs removal suggests someone should correct things. I think the top candidates for introducing bad fields are the guess methods and pragmas both of which would probably benefit from correction. Maybe a logger warning here or on results.errors? Maybe put it behind a verbose option? Thoughts?

setup.py
```diff
@@ -39,7 +39,7 @@
     long_description=readme,
     entry_points = {
         'console_scripts': [
-            'pytmc = pytmc.bin.makerecord:main',
+            'pytmc = pytmc.bin.pytmc:main',
```
n-wbrown: 👍

pytmc/xml_collector.py
klauer: So, addressing @teddyrendahl's comment above and yours here:

The current behavior with the linter is to show a message about invalid fields and other warnings/errors. In the event of one or more errors, the `pytmc` CLI tool will not output a db file, exiting with status code `1`. The user can force a file to be generated in this case with an additional flag. @teddyrendahl suggests that it should output a file regardless, while my gut feeling is that it should not. 

We should get a consensus on this before moving forward.

klauer:
@teddyrendahl, with the user being able to add fields arbitrarily - even with the template refactor we've been talking about - we will likely be seeing many errors that could be caught prior to IOC startup. That is what this PR is for. 

To date, most of the provided `.tmc` files in pytmc have invalid fields or other errors. Additionally, the first "production" usage of pytmc in https://github.com/slaclab/pytmc/issues/36 uncovered a whole host of issues. I don't much see the need for justifying this any further...

pytmc/xml_collector.py
teddyrendahl: I yield to @klauer's point that you don't want to create a file you don't want being used. I'm fine not emitting if there is a detailed error message containing the record name

klauer:
pytmc script now shows context around errors for easier debugging:
```sh
$ pytmc tests/tmc_files/invalid/generic_w_pragmas7.tmc  test.db --dbd tests/ads.dbd
ERROR:pytmc.bin.pytmc:[<string> line 42] recordtype 'ao' has not field 'ONAM'
ERROR:pytmc.bin.pytmc:   [db:41] record(ao,"TEST:MAIN:ITERATOR:VALUE"){
ERROR:pytmc.bin.pytmc:   [db:42]   field(PINI, "1")
ERROR:pytmc.bin.pytmc:   [db:43]   field(TSE, -2)
ERROR:pytmc.bin.pytmc:[<string> line 115] recordtype 'ai' has not field 'OUT'
ERROR:pytmc.bin.pytmc:   [db:110] record(ai,"TEST:SAMPLE_GVL:TEST_GLOBAL"){
ERROR:pytmc.bin.pytmc:   [db:111]   field(DTYP, asynFloat64)
ERROR:pytmc.bin.pytmc:   [db:112]   field(PINI, "1")
ERROR:pytmc.bin.pytmc:   [db:113]   field(TSE, -2)
ERROR:pytmc.bin.pytmc:   [db:114]   field(SCAN, "Passive")
ERROR:pytmc.bin.pytmc:   [db:115]   field(PREC, "3")
ERROR:pytmc.bin.pytmc:   [db:116] }
ERROR:pytmc.bin.pytmc:[<string> line 121] recordtype 'ao' has not field 'ONAM'
ERROR:pytmc.bin.pytmc:   [db:118] record(ao,"TEST:MAIN:@2:VALUE"){
ERROR:pytmc.bin.pytmc:   [db:119]   field(PINI, "1")
ERROR:pytmc.bin.pytmc:   [db:120]   field(TSE, -2)
ERROR:pytmc.bin.pytmc:   [db:121]   field(DTYP, "asynInt32")
ERROR:pytmc.bin.pytmc:   [db:122]   field(OUT, "@asyn($(PORT),0,1)ADSPORT=851/MAIN.second_test_iterator.value=")
ERROR:pytmc.bin.pytmc:Linter errors - failed to create database. To disable this behavior, use the flag --allow-errors
```

n-wbrown:
Looks good, thanks everyone!

PR #44, opened 2019-04-15
CONDA Deployment

teddyrendahl:
I'd like to get this is in so we can start deploying this on `pcds-tag` this will be necessary to include it as part of our standard Python releases

* Setups a CONDA recipe
* Move a few packages around between `requirements` and `dev-requirements`
* Remove the version specifications on the packages
* Use CONDA to build prior to tests on Travis
* Puts machinery in Travis to deploy built packages to `pcds-tag` and `pcds-dev`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-Jinja2==2.10
+Jinja2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,1 @@
 Jinja2==2.10
-pytest
-Sphinx==1.6.5
-sphinx-rtd-theme==0.2.4
-sphinxcontrib-websupport==1.0.1
-sphinx_autodoc_typehints==1.3.0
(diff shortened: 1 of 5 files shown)


n-wbrown:
@teddyrendahl Do you think it would make sense to use the latest version of Jinja as well rather than using a hard specifier? If so we might as well give it the same treatment as the `dev-requirements.txt`. Everything else looks great!

requirements.txt
```diff
@@ -1,7 +1 @@
-Jinja2==2.10
-pytest
-Sphinx==1.6.5
-sphinx-rtd-theme==0.2.4
-sphinxcontrib-websupport==1.0.1
-sphinx_autodoc_typehints==1.3.0
-versioneer==0.18
+Jinja2
```
teddyrendahl: Nice catch @n-wbrown!

PR #46, opened 2019-04-16
REF: add _skip_if_field_set decorator

klauer:
Split methods that guess multiple fields for simplicity.
Removes a fair amount of boilerplate try/except in the guess methods.

### DIFF
--- a/pytmc/xml_collector.py
+++ b/pytmc/xml_collector.py
@@ -5,6 +5,7 @@ This file contains the objects for intaking TMC files and generating python
 interpretations. Db Files can be produced from the interpretation
 """
 import logging
+import functools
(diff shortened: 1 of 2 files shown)


klauer:
@teddyrendahl There may be a field or two where that's the case, but I can't think of one off the top of my head... Most of this is based on some heuristics of what's in the `.tmc`.

pytmc/xml_collector.py
```diff
@@ -764,18 +767,43 @@ def generate_files(records):
         """
         raise NotImplementedError
 
-    def guess_common(self):
+    def _skip_if_field_set(field):
```
n-wbrown: I like this decorator!

PR #48, opened 2019-04-16
ADS Port

teddyrendahl:
The ADS Port can be found in the `tmc` file under `Modules/Module/Properties`. We can not assume that the ADS port will always be the default `851` because certain deployments may have multiple PLCs projects running in one solution, this means these variables will be accessed via different ports.

This PR grabs the ADS port from the TMC file and passes it into each `BaseRecordPackage`. I made `ads_port` are required argument, but this could be relaxed into an optional keyword with a default of `851` if people think that is appropriate and do not want to break API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 # Custom
 *.swp
 *.swo
-
+tests/tmc_files/*.db
+tests/tmc_files/invalid/*.db
 
 
 .DS_store
--- a/pytmc/xml_collector.py
+++ b/pytmc/xml_collector.py
@@ -344,7 +344,7 @@ class TmcFile:
         packages. requires self.all_singular_TmcChains to be populated.
         """
         for singular_chain in self.all_singular_TmcChains:
-            brp = BaseRecordPackage(chain=singular_chain)
+            brp = BaseRecordPackage(self.ads_port, chain=singular_chain)
             #brp = BaseRecordPackage(chain=singular_chain, origin=chain)
             self.all_RecordPackages.append(brp)
 
@@ -583,7 +583,7 @@ class BaseRecordPackage:
     _required_keys = {'pv', 'type', 'field'}
     _required_fields = {'DTYP', }
 
-    def __init__(self, chain=None, origin=None):
+    def __init__(self, ads_port, chain=None, origin=None):
         """
         All subclasses should use super on their init method.
         """
@@ -630,7 +630,7 @@ class BaseRecordPackage:
             "asyn_standard_file.jinja2"
         )
 
-        self.ads_port = 851
+        self.ads_port = ads_port
 
     def apply_config_validation(self):
         """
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,7 +1,8 @@
 # Custom
 *.swp
 *.swo
-
+tests/tmc_files/*.db
+tests/tmc_files/invalid/*.db
```
klauer: These are my fault. That said, I think the tests generating these and allowing inspection of them is somewhat useful. We can remove them if desirable (@n-wbrown?)

pytmc/xml_collector.py
```diff
@@ -119,6 +120,19 @@ def __init__(self, filename):
...
+    def ads_port(self):
+        """Return the ADS Port defined in the TMC file under ApplicationName"""
+        if not self.root:
+            return None
+        # Grab the ApplicationName from XML Tree
+        element = self.root.find("./Modules/Module/Properties/"
+                                 "Property/[Name='ApplicationName']")
+        element_value = element.find('./Value').text
+        # Parse the value for the port number and convert
+        port = int(re.search('(\d+)', element_value).group())
```
klauer: What does this string look like in the `.tmc` file, just out of curiosity?

pytmc/xml_collector.py
```diff
@@ -119,6 +120,19 @@ def __init__(self, filename):
             "asyn_standard_file.jinja2"
         )
 
+    @property
+    def ads_port(self):
+        """Return the ADS Port defined in the TMC file under ApplicationName"""
```
n-wbrown:
```suggestion
        """
        Returns
        -------
        int
            The ADS Port defined in the TMC file under ApplicationName
        """
```

pytmc/xml_collector.py
teddyrendahl: `Port_851`

.gitignore
teddyrendahl: Up to the group at large. I'm indifferent.

n-wbrown: Notes from our conversation: having the generated .db is great for testing/debugging but shouldn't be pushed to github. These additions to the `.gitignore` are good.

tests/test_xml_collector.py
```diff
@@ -610,21 +615,21 @@ def test_BaseRecordPackage_render_record():
...
+    brp = BaseRecordPackage(851)
     brp.cfg.add_config_line('pv', 'example_pv')
     brp.cfg.add_config_line('type', 'motor')
     brp.cfg.add_config_field("PINI", "1")
     assert brp.ID_type() == 'motor'
 
 
-def test_BaseRecordPackage_guess_PINI():
-    brp = BaseRecordPackage()
+def test_BaseRecordPackage_guess_common():
```
n-wbrown:
```suggestion
def test_BaseRecordPackage_guess_PINI():
```

n-wbrown: Undo mistaken change of test name

teddyrendahl:
**This branch was heavily rebased to solve merge conflicts with multiple concurrent PRs**

n-wbrown:
Thanks @teddyrendahl.

PR #75, opened 2019-04-26
ENH: add pytmc-debug GUI tool for inspecting pytmc processing

klauer:
image

Although this script has a dependency on `qtpy` (and therefore `PyQt5`/`Qt5`/etc), I think it can be considered optional. I had intended to make a command line-only alternative to the Qt version, but I'm not sure it's worthwhile.

### DIFF
--- a/pytmc/bin/pytmc.py
+++ b/pytmc/bin/pytmc.py
@@ -14,19 +14,20 @@ logger = logging.getLogger(__name__)
 description = __doc__
 
 
-def make_db_text(tmc_file_obj, *, dbd_file=None, allow_errors=False,
-                 show_error_context=True):
+class LinterError(Exception):
+    ...
+
+
+def process(tmc_file_obj, *, dbd_file=None, allow_errors=False,
+            show_error_context=True):
     '''
-    Create an EPICS database from a TmcFile
+    Process a TmcFile, readying it for database generation
 
     Parameters
     ----------
     tmc_file_obj : TmcFile
 
-    Returns
-    -------
-    dbtext : str
-        The rendered database text
+
     '''
     tmc_file_obj.create_chains()
     tmc_file_obj.isolate_chains()
@@ -62,10 +63,9 @@ def make_db_text(tmc_file_obj, *, dbd_file=None, allow_errors=False,
             logger.error('Linter errors - failed to create database. '
                          'To disable this behavior, use the flag '
                          '--allow-errors')
-            sys.exit(1)
+            raise LinterError('Failed to create database')
 
-    return tmc_file_obj.render()
-
+    return tmc_file_obj
 
 
 def main():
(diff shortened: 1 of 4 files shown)


n-wbrown:
1. This is really cool! 
2. Do we want to add a second requirements file for this new tool (e.g. debug-gui-requirements.txt)?

klauer:
1. Thanks, @n-wbrown! Think it's the best we can do as far as representing the chains/records/config information?

2. Hmm, maybe it could be added to `dev-requirements`?

n-wbrown:
1. I think the only recommendations I could make are pretty minor. Maybe numbering the rows, labeling the individual windows, or indenting successively encapsulated layers? I find the interface is completely readable for a pytmc developer as is so I'm having a hard time recommending any real changes. If we intend to expand its usage so any TwinCAT developer can use it (which now that I'm thinking about it seems like it might be a great PLC debugging/development tool), then we might start making these small readability upgrades. If we start doing this, it could also be neat to feed the output of the linter into `pytmc_debug` and highlight where the issues originated. 

2. That seems appropriate.

pytmc/bin/pytmc.py
```diff
@@ -14,19 +14,20 @@
 description = __doc__
 
 
-def make_db_text(tmc_file_obj, *, dbd_file=None, allow_errors=False,
-                 show_error_context=True):
+class LinterError(Exception):
+    ...
```
teddyrendahl: Educate me. Why not `pass`?

pytmc/bin/pytmc_debug.py
```diff
@@ -0,0 +1,283 @@
...
+    item_selected = Signal(object)
+
+    def __init__(self, tmc, dbd):
+        super().__init__()
+        self.tmc = tmc
+        self.chains = {}
+        self.records = {}
+
+        for record in tmc.all_RecordPackages:
+            record_text = record.render_record()
```
teddyrendahl: Think you might want to include this in the `try / except` block. `render_record` could raise some errors if we have an invalid `RecordPackage`

pytmc/bin/pytmc_debug.py
```diff
@@ -0,0 +1,283 @@
...
+    dbd : DbdFile, optional
+        The dbd file to lint against
+    '''
+    app = QtWidgets.QApplication([])
+    interface = TmcSummary(tmc, dbd)
+    interface.show()
+    sys.exit(app.exec_())
+
+
+def main():
```
teddyrendahl: Might want to also add a hook for verbosity of `logging` stream?

pytmc/bin/pytmc.py
klauer: Either are equally valid, assuming 3.x+; so it's just a stylistic choice.

pytmc/bin/pytmc_debug.py
klauer: I had it and thought it just added unnecessary spam during startup, when what I wanted to see was the GUI. If either you or @n-wbrown feel strongly about it, though, I'd be happy to add it back.

teddyrendahl: As in you had the `DEBUG` logging stream? You could just set it to `WARNING` unless otherwise specified?

klauer: Sorry - I just meant that this tool had the same `--log` flag as in the main `pytmc` utility. Added it back in 12a954c259648f884132f634a9433d2b4bf42517.

PR #78, opened 2019-05-01
REF:  BaseRecordPackage Refactor

### LINKED ISSUE #64: Decide on pattern for setpoint/readback records
With asyn [R4-26](https://epics.anl.gov/modules/soft/asyn/R4-26/RELEASE_NOTES.html), a new option to allow a record to act as both a setpoint and readback was introduced:

> Added capability for output records to update their values with a callback from the driver. 
> This is enabled by adding the following info tag to the output record:
> `info(asyn:READBACK, "1")`
> The output record value will update any time the driver does a callback with a new value. 

pytmc currently uses this, as it only requires generating one record for a specific chain.

While convenient for record generation, the above goes against the pattern used almost everywhere else in the EPICS world, where there are distinct records for setpoint and their associated readbacks (`_RBV`). 

Mark, as the primary author of asyn, outlines rather nicely some arguments against using the new `asyn:READBACK` functionality in this tech-talk post: https://epics.anl.gov/tech-talk/2016/msg00157.php .

We should decide which style we're going to use going forward and stick with it consistently. 

At least @n-wbrown, @zllentz, @teddyrendahl should weigh in before we proceed.
My vote is on separating setpoint/readback, though it will be more work.

### LINKED ISSUE #67: Allow Jinja templates to be used instead of guesses
Needed for custom record types such as "motor". Will be able to be sourced from outside `pytmc` via environment variable

### LINKED ISSUE #69: Split BaseRecordPackage
Strip down `BaseRecordPackage` to base class. Put current logic from `BaseRecordPackage` into a new class for basic TwinCat types. `TwincatTypeRecordPackage`

teddyrendahl:
# Description
This rips apart `BaseRecordPackage` along with the guessing mechanics in favor of a more straight forward schema. The main motivation is that we might want a single Twincat variable to spawn multiple records so the `RecordPackage` infrastructure needed to accommodate this change.

`EpicsRecord`: Load your fields and it handles filling an arbitrary Jinja template
`RecordPackage`: Stub for all other subclasses
`TwincatTypeRecordPackage`: A record package derived from a builtin Twincat type. 
 
We then have subclasses of `TwincatTypeRecordPackage`as needed to support all the types. Finally, `TwincatTypeRecordPackage` looks at the type of the last element in the chain and picks the appropriate subclass for you. In this way it functions very much like the guessing mechanics formerly in `BaseRecordPackage`

### Changes worth noting
1. `SCAN` now is set to `I/O Intr` for input records, ignored for output
2. We now never write back to the `Configuration` object with our "guesses". That object only will hold what was given to us via reading the TmC file. The sticking point was what do we write back if we are creating multiple records?
3. With respect to the point above, the internal validation of the record still exists but now will only check with the required tags and fields that we need from the user (which on the base implementation is only the pvname)
4. For strings, `NELM` is fixed at 81. In the previous implementation if a larger string would have been written to the twincat variable it would be truncated based on the starting variable value.
5. We longer support someone manually entering the "type" field as an EPICS RTYP. If you give us a standard Twincat type we will always infer which EPICS record type to use based on that. We many repurpose this later to support motors e.t.c
6. I removed a lot of the interface that was not being used and marked as `NotImplemented`
7. We support `io` and `i` tags,  anything labeled as just `o` will still have a readback created for it.
8. All readback records get `_RBV` added on to the pvname if it does not exist.

# Motivation
Closes #69 
Closes #67 
Closes #64 

# Tests
I decided to keep the tests as much as possible to prove to myself I didn't break anything. If this PR is accepted I would like to write a parallel test suite that more explicitly uses the new API.

# Documentation
Will add once we agree this is the route we want to go on...


pytmc/record.py
```diff
@@ -0,0 +1,28 @@
+"""Individual EPICS Database Record creation"""
+from jinja2 import Environment, PackageLoader
+
+
+class EPICSRecord:
+    """Representation of a single EPICS Record"""
+    def __init__(self, pvname, record_type, fields=None, template=None):
+        self.pvname = pvname
+        self.record_type = record_type
+        self.fields = dict(fields) or dict()
```
klauer: This won't work as you're expecting:
```python
In [1]: dict(None) or dict()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-149716818289> in <module>()
----> 1 dict(None) or dict()

TypeError: 'NoneType' object is not iterable
```

```suggestion
        self.fields = dict(fields) if fields is not None else {}
```

pytmc/record.py
```diff
@@ -0,0 +1,28 @@
...
+        self.record_template = self.jinja_env.get_template(
+            self.template
+        )
+
+    def render_template(self):
+        """Render the provided template"""
+        return self.record_template.render(record=self)
+
+    def __repr__(self):
+        return f"EPICSRecord ({self.pvname}, RTYP={self.record_type})"
```
klauer:
```suggestion
        return f"EPICSRecord({self.pvname!r}, record_type={self.record_type!r})"
```

pytmc/templates/asyn_standard_record.jinja2
```diff
@@ -1,8 +1,6 @@
-record({{type}},"{{pv}}"){
-{% if info %}
-  info(asyn:READBACK,"1")
-{% endif %}
-{% for f in field%}
-  field({{f}}, {{field[f]}})
+record({{record.record_type}},"{{record.pvname}}"){
```
klauer:
```suggestion
record({{record.record_type}}, "{{record.pvname}}"){
```

pytmc/templates/asyn_standard_record.jinja2
```diff
@@ -1,8 +1,6 @@
...
+{% block add_fields  %}{% endblock %}
+{% for f in record.fields%}
+  field({{f}}, {{record.fields[f]}})
```
klauer: Reminder that we should refactor quoting - requiring none in `EPICSRecord` and automatically adding quotes here. Can be done outside of this PR.

pytmc/xml_collector.py
```diff
@@ -422,7 +432,7 @@ def render(self):
         """
         rec_list = []
         for pack in self.all_RecordPackages:
-            record_str = pack.render_record()
+            record_str = pack.render_records()
```
klauer: This remains a bit inconsistent, naming-wise. Every render method should be `render_<something>()` or just plain `render()`.

pytmc/xml_collector.py
```diff
@@ -536,7 +546,8 @@ def _recursive_permute(self, master_list, so_far=None):
                 extras.append(deepcopy(chain))
         so_far.extend(extras)
 
-        for term, term_index in zip(master_list[0], range(len(master_list[0]))):
+        for term, term_index in zip(master_list[0],
```
klauer:
```suggestion
        for term_index, term in enumerate(master_list[0]):
```

pytmc/xml_collector.py
```diff
@@ -874,72 +736,122 @@ def valid(self):
                                   for key in self._required_fields)
         return has_required_keys and has_required_fields
 
-    def render_record(self):
+    @property
+    def records(self):
+        """Generated :class:`.EPICSRecord` objects"""
+        return NotImplemented
```
klauer:
```suggestion
        raise NotImplementedError()
```

Returning `NotImplemented` should be just for `__lt__` et al (See https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror)

pytmc/xml_collector.py
```diff
@@ -874,72 +736,122 @@ def valid(self):
...
         string
-            Jinja rendered entry for this BaseRecordPackage
+            Jinja rendered entry for the RecordPackage
         """
-        simple_dict = self.cfg_as_dict()
         if not self.valid:
-            logger.error('Unable to render record: %s', simple_dict)
+            logger.error('Unable to render record: %s', self)
             return
+        return '\n'.join([record.render_template()
```
klauer: Maybe?

```suggestion
        return '\n\n'.join([record.render_template().strip()
```

That is, ignore any whitespace in the template at the beginning or end, and ensure two line breaks between each record.

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
-
-    def render_to_string(self):
-        """
-        Create the individual record to be inserted in the DB file. To be
-        returned as a string.
-        redundant?
-        """
-        raise NotImplementedError
+
+data_types = {
```
klauer: I like to use `@register_record_type('BOOL')`-type class decorators for these sorts of registries. It keeps the mapping closer to where one might see it while reading the source, with less jumping around.

e.g., you would decorate `BinaryRecordPackage` as follows:
```python
@register_record_type('BOOL')
class BinaryRecordPackage(TwincatTypeRecordPackage):
    ...
```

Completely up to you, though, as this is just fine.

pytmc/xml_collector.py
```diff
@@ -874,72 +736,122 @@ def valid(self):
...
+class TwincatTypeRecordPackage(RecordPackage):
+    """
+    The common parent for all RecordPackages for basic Twincat types
+
+    This main purpose of this class is to handle the parsing of the pragma
+    chains that will be shared among all variable types. This includes setting
+    the appropriate asyn port together and handling the "io" directionality. If
+    you have a :class:`.TmcChain` but are not certain which class is
+    appropriate use the :meth:`.from_chain` class constructor and the correct
+    subclass will be chosen based on the given variable information
```
klauer:
```suggestion
    subclass will be chosen based on the given variable information.
```

pytmc/xml_collector.py
```diff
@@ -874,72 +736,122 @@ def valid(self):
...
+
+    In order to subclass:
```
klauer:
```suggestion
    In order to subclass::
```

I think then the next block will have to be indented (check the sphinx output).
Also, these classes need to be added to the autoclass lists in the docs.

pytmc/xml_collector.py
```diff
@@ -874,72 +736,122 @@ def valid(self):
...
+    output_rtyp = NotImplemented
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # use form list of dicts,1st list has 1 entry per requirement
+        # dict has form {'path':[],'target':n} where n is any variable
+        self.validation_list = None
+
+    def __init_subclass__(cls, **kwargs):
+        """Magic to have field_defaults be the combination of hierarchy"""
```
klauer: 👍 nice magic!

pytmc/xml_collector.py
```diff
@@ -874,72 +736,122 @@ def valid(self):
...
-        ----------
-        records : list
-            list of all incoming Record objects
+        Apply the guessing module. Assert that the proper fields exist.
 
         Returns
         -------
-        str
-            The Jinja-made output of the full resulting file.
+        List
```
klauer:
```suggestion
        list
```

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
-            ty, = self.cfg.get_config_lines('type')
-            return False
-        except ValueError:
-            pass
-
-        try:
-            direction = self.io_direction
+            io, = self.cfg.get_config_lines('io')
+            io = io['tag']
+            if 'o' in io:
```
klauer: 👍 simplification

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
+        record.fields['INP'] = self._asyn_port_spec + '?"'
+        record.fields['DTYP'] = self.dtyp
+        record.fields['SCAN'] = '"I/O Intr"'
 
-        The following is taken from the EPICS wiki: "This field specifies the
-        device type for the record. Each record type has its own set of device
-        support routines which are specified in devSup.ASCII. If a record type
-        does not have any associated device support, DTYP and DSET are
-        meaningless."
+        # Update with given pragamas
```
klauer:
```suggestion
        # Update with given pragmas
```

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
-        elif 'i' in io:
-            assign_symbol = "?"
-            field_type = 'INP'
-        elif 'o' in io:
-            assign_symbol = "="
-            if self.chain.last.is_array or self.chain.last.is_str:
-                field_type = 'INP'
-            else:
-                field_type = 'OUT'
+        # Update with given pragamas
```
klauer:
```suggestion
        # Update with given pragmas
```

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
-    @_skip_if_field_set('SCAN')
-    def guess_SCAN(self):
-        """
-        add field for SCAN field
+class BinaryRecordPackage(TwincatTypeRecordPackage):
+    """Create a set of records for a binary Twincat Variable"""
+    input_rtyp = 'bi'
+    output_rtyp = 'bo'
+    dtyp = '"asynInt32"'
+    field_defaults = {'ZNAM': '"Zero"', 'ONAM': '"One"'}
```
klauer:
```suggestion
    field_defaults = {'ZNAM': '"Zero"', 'ONAM': '"One"'}
```

@n-wbrown objections to changing this to "Off", "On"? I find a mapping of 0=Zero and 1=One to be not so helpful.

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
-        -------
-        bool
-            Return a boolean that is true iff a change has been made.
-        """
-        if self.chain.last.tc_type in {'BOOL'}:
-            self.cfg.add_config_field("ONAM", "One")
-            return True
-        return False
+class EnumRecordPackage(TwincatTypeRecordPackage):
+    """Create a set of record for a floating point Twincat Variable"""
```
klauer:
```suggestion
    """Create a set of record for an ENUM Twincat Variable"""
```

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
-        Returns
-        -------
-        bool
-            Return a boolean that is true iff a change has been made.
-        """
-        if self.chain.last.tc_type in {'BOOL'}:
-            self.cfg.add_config_field("ZNAM", "Zero")
-            return True
+class WaveformRecordPackage(TwincatTypeRecordPackage):
+    input_rtyp = 'waveform'
```
klauer: No docstring

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
+                      'USINT': '"asynInt8',
+
+                      'WORD': '"asynInt16',
+                      'INT': '"asynInt16',
+                      'UINT': '"asynInt16',
+
+                      'DWORD': '"asynInt32',
+                      'DINT': '"asynInt32',
+                      'UDINT': '"asynInt32',
+                      'ENUM': '"asynInt16',  # -> Int32?
```
klauer: Another reminder this needs reviewing (what are ENUMs actually, in twincat... ?)

pytmc/xml_collector.py
```diff
@@ -956,308 +868,222 @@ def io_direction(self):
...
+        # Waveform records only have INP fields!
+        record.fields['INP'] = record.fields.pop('OUT')
+        return record
+
+
+class StringRecordPackage(TwincatTypeRecordPackage):
+    """RecordPackage for broadcasting string values"""
+    input_rtyp = 'waveform'
+    output_rtyp = 'waveform'
+    dtyp = '"asynInt8'
```
klauer: I think `asynOctetRead` may be more accurate here...

tests/test_linter.py
```diff
@@ -31,3 +24,19 @@ def test_db_linting(dbd_file, tmp_path, filename):
 
     results = tmc_file.validate_with_dbd(dbd_file)
     assert results.success
+
+
+@pytest.mark.parametrize(
+        'filename',
+        conftest.INVALID_TMC_FILES,
+        ids=[f.name for f in conftest.INVALID_TMC_FILES])
+def test_invalid_db_linting(dbd_file, tmp_path, filename):
```
klauer: Better, but as a reminder, we should nuke these obsolete tmc files from the repo altogether.

tests/test_record.py
```diff
@@ -0,0 +1,17 @@
...
+                         'ONAM': 'In'}}
+
+    ec = EPICSRecord(**kwargs)
+    record = ec.render_template()
+    print(record)  # For debug purposes
+    assert kwargs['pvname'] in record
+    assert kwargs['record_type'] in record
+    for key, value in kwargs['fields'].items():
+        assert key in record
+        assert value in record
```
klauer: Maybe lint the record at the end?

pytmc/templates/asyn_standard_record.jinja2
teddyrendahl: Planned as my next PR, but this one seemed large enough 😉

pytmc/xml_collector.py
teddyrendahl: Good point. I'm open to changes in the API. Kind of like just straight `render`. Thoughts @n-wbrown?

tests/test_linter.py
teddyrendahl: 👍

tests/test_record.py
teddyrendahl: This won't be a valid EPICS record. It is up to the record packages to configure all the default fields appropriately

pytmc/xml_collector.py
teddyrendahl: Might just put this in a separate issue if you don't mind?

klauer: :+1:

tests/test_record.py
klauer: I guess my intention was more to say "create a basic EPICS record and verify that it's valid" - as a simple check of `pytmc.EPICSRecord` on its own. Just a thought, in any case.

pytmc/xml_collector.py
teddyrendahl: Issue #79

pytmc/xml_collector.py
teddyrendahl: If I had thought of this before then I would implement it this way. As it is I don't think it is worth going back unless these subclasses get complex enough that they start being in different files (which is plausible)

pytmc/xml_collector.py
teddyrendahl: I don't think the indent did what you expected, added classes to docs.

tests/test_record.py
teddyrendahl: You have resisted my attempt at laziness, added in f4cc2a1

teddyrendahl:
With only one comment left to resolve (and could easily be changed later), I think this is ready for a final review and approval.

We should still make sure that we create a new tag and release before this is merged. I can do this if you are too busy @n-wbrown

pytmc/xml_collector.py
n-wbrown: Only minor ones. Given that this is just a default, I was thinking we should try to presume the least about our hypothetical user's use-case and try to leave things as literal as possible. I'm concerned we might be over interpreting intent especially given that this is just a bool. If you're strongly in favor of "On"/"Off" I'd say let's change it to "On"/"Off".

PR #83, opened 2019-06-19
PERF: reduce amount of deepcopy calls in build_singular_chains

klauer:
This appears to be at the core of https://github.com/slaclab/pytmc/issues/82.

It was not an infinite loop as I had guessed, but it was taking on the order of minutes and gigabytes of memory due to unnecessary deepcopy calls.

With the current master, I gave up after 2 minutes of waiting:

image

With this PR, it goes at a not-so-unreasonable speed:

```sh
$ time pytmc --allow-errors VonHamos01.tmc test.db

real    0m9.965s
user    0m9.776s
sys     0m0.157s
```

### DIFF
--- a/pytmc/xml_collector.py
+++ b/pytmc/xml_collector.py
@@ -566,21 +566,20 @@ class TmcChain:
             List of TmcChains. Each chain is bound to one of the possible paths
             given the configurations available at each step.
         """
-        name_sequences = self._recursive_permute(self.forkmap())
+        forkmap = self.forkmap()
+        name_sequences = self._recursive_permute(forkmap)
         logging.debug(str(name_sequences))
+        if any(config.pragma is None for config in self.chain):
+            return []
+
         results = []
-        append = True
+
         for seq in name_sequences:
             new_TmcChain = deepcopy(self)
+            results.append(new_TmcChain)
+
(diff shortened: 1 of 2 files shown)


klauer:
Surprisingly, caching the name with what I believe the intended mechanism is (in fda21c0) shaved a few seconds off: 

```
real    0m6.622s
user    0m6.401s
sys     0m0.173s
```

And another second off in not unnecessarily calling `str(entry)` in a68b549: 

```
real    0m5.345s
user    0m5.133s
sys     0m0.171s
```

pytmc/xml_collector.py
```diff
@@ -566,21 +566,20 @@ def build_singular_chains(self):
             List of TmcChains. Each chain is bound to one of the possible paths
             given the configurations available at each step.
         """
-        name_sequences = self._recursive_permute(self.forkmap())
+        forkmap = self.forkmap()
+        name_sequences = self._recursive_permute(forkmap)
         logging.debug(str(name_sequences))
+        if any(config.pragma is None for config in self.chain):
```
klauer: This isn't quite correct. The `zip` from the removed code means that the chain.pragma should only be checked up to `self.chain[:len(seq)]`.

klauer:
Tests are apparently not checking what I mentioned above; do not merge yet.

PR #86, opened 2019-07-10
CI: fix docs

klauer:
One straggler commit I forgot to push as part of #85, plus (hopefully) a fix to building/publishing docs.


klauer:
Self-merging as these are minor, and docs are building now.

PR #89, opened 2019-07-24
REF: command line to make optional deps truly optional

### LINKED ISSUE #87: Top-level script `pytmc` makes optional dependencies required
As `pytmc-debug`, for example, imports `qtpy`. 

Temporary workaround: `pytmc-stcmd` and such work on their own.

klauer:
Also refactors such that testing can be done more effectively, and tools more easily usable by Python (interpreting kwargs and not argparse namespaces).

* Removes ability to use `pytmc-db` et al, pushing users toward the subcommand usage: `pytmc db`.

Closes #87


dev-requirements.txt
```diff
@@ -1,10 +1,13 @@
...
+pytest-qt
+qtpy
 sphinx
-sphinx-rtd-theme
 sphinx-argparse
-sphinxcontrib-websupport
+sphinx-rtd-theme
 sphinx_autodoc_typehints
-flake8
+sphinxcontrib-websupport
```
teddyrendahl: Interested in what these extra Sphinx plugins do...

pytmc/bin/pytmc.py
```diff
@@ -64,8 +60,17 @@ def main():
         sub.set_defaults(func=main)
 
     args = top_parser.parse_args()
+    kwargs = vars(args)
+    log_level = kwargs.pop('log_level')
+
+    logger = logging.getLogger('pytmc')
+    logger.setLevel(log_level)
+    logging.basicConfig()
```
teddyrendahl: I get so confused here but I think you need to make sure that you set the logging level of whatever `logging.Handler` gets created in `basicConfig` as well? I typically just pass it into the constructor but could be wrong here.

pytmc/bin/pytmc.py
```diff
@@ -3,43 +3,40 @@
...
+def _try_import(module):
+    relative_module = f'.{module}'
+    return importlib.import_module(relative_module, 'pytmc.bin')
+
+
+def _build_commands():
+    global DESCRIPTION
+    result = {}
+    for module in sorted(MODULES):
+        mod = _try_import(module)
```
teddyrendahl: Am I missing where this is optional? A failed import of one of the sub commands will still crash everything?

pytmc/bin/pytmc.py
```diff
@@ -3,43 +3,40 @@
...
+        try:
+            result[module] = (mod.build_arg_parser, mod.main)
+        except AttributeError:
+            continue
```
teddyrendahl: Not sure what this `AttributeError` is really watching for, but might be nice to have a message about certain commands being unavailable due to dependency issues?

dev-requirements.txt
klauer: Oops, looks like I had accidentally sorted the dev-requirements. The actual changes here are requiring qt things for the tests. 

Not sure about that `websupport` one.

pytmc/bin/pytmc.py
klauer: That was the entire point of the PR and I missed it. Thanks for paying close attention 👍

pytmc/bin/pytmc.py
klauer: I think it's probably vestigial - will remove.

pytmc/bin/pytmc.py
klauer: I think caproto is probably a rather good model of logging practices after the recent-ish refactor:
https://github.com/caproto/caproto/blob/3b802c885b4ecb3c9aa70947b86dddbb6aded2c4/caproto/_log.py#L340

We really should be adding our own handler and setting the level on that. The code as-is does work though 🤷‍♂

pytmc/bin/pytmc.py
```diff
@@ -3,43 +3,40 @@
...
+
+
+def _build_commands():
+    global DESCRIPTION
+    result = {}
+    for module in sorted(MODULES):
+        try:
+            mod = _try_import(module)
+        except ImportError:
+            ...
```
teddyrendahl:
```suggestion
            DESCRIPTION += f'\n    Command line tool unavailable due to dependency isssues - pytmc {module}'
```

klauer: :+1: I like that pattern. I moved it to the bottom of the description, though, so it doesn't get in the way:

```
$ pytmc --help
usage: pytmc [-h] [--log LOG_LEVEL]
             {db,debug,stcmd,summary,types,xmltranslate} ...

`pytmc` is the top-level command for accessing various subcommands.

Try::

    $ pytmc db --help
    $ pytmc debug --help
    $ pytmc stcmd --help
    $ pytmc summary --help
    $ pytmc types --help
    $ pytmc xmltranslate --help

WARNING: pytmc 'fubar' is unavailable due to:
        ModuleNotFoundError: No module named 'pytmc.bin.fubar'

positional arguments:
  {db,debug,stcmd,summary,types,xmltranslate}
                        Possible subcommands

optional arguments:
  -h, --help            show this help message and exit
  --log LOG_LEVEL, -l LOG_LEVEL
                        Python logging level (e.g. DEBUG, INFO, WARNING)
```

klauer:
Agreed on the coverage concerns, @teddyrendahl. Looking into it locally:
image
(previous results vs current)

It seems like the addition of the submodules in the coverage report was enough to bring it down?

teddyrendahl:
I believe you might just be being punished for testing files that were not imported below. I'm fine with merging as is.

PR #90, opened 2019-07-24
FIX: AMS ID is on the TopLevelProject

klauer:
cc @zllentz - as we ran into this yesterday.

Output from the project during the demo:
```
$ pytmc summary --plcs tests/projects/motion-demo/motion-demo/motion-demo/motion-demo.tsproj
PLC Project (1): Demo1
======================

    Project path: /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo1/Demo1.plcproj
    TMC path:     /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo1/Demo1.tmc
    AMS ID:       5.57.239.146.1.1
    IP Address:   5.57.239.146 (* based on AMS ID)
    Port:         851

    Source files:
        1.) /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo1/PlcTask.TcTTO
        2.) /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo1/POUs/Main.TcPOU

    POUs:
        1.) Main

PLC Project (2): Demo2
======================

    Project path: /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo2/Demo2.plcproj
    TMC path:     /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo2/Demo2.tmc
    AMS ID:       5.57.239.146.1.1
    IP Address:   5.57.239.146 (* based on AMS ID)
    Port:         852

    Source files:
        1.) /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo2/PlcTask.TcTTO
        2.) /Users/klauer/docs/Repos/pytmc/tests/projects/motion-demo/motion-demo/motion-demo/Demo2/POUs/Main.TcPOU

    POUs:
        1.) Main
```

PR #92, opened 2019-07-25
ENH: pytmc-iocboot

klauer:
Couples with pcdshub/ads-ioc#7

General idea is the following:

```bash
$ pytmc iocboot /path/to/plc.tsproj /path/to/ads-ioc/iocBoot/templates/
# Creates directories: ioc-plc-name1, ioc-plc-name2, ...
# Creates Makefiles: ioc-plc-name/Makefile

$ cd ioc-plc-name
$ vim Makefile
# Customize - if necessary - to add more db files or change pytmc stcmd options

$ make
# Shells out to pytmc stcmd;
# Creates st.cmd, ioc-plc-name.db, envPaths

$ ./st.cmd
# Runs IOC
```

Remaining:
- [x] Binary location
- [x] envPaths generation / copying?
- [ ] Archiver request files (maybe)

### DIFF
--- a/pytmc/bin/stcmd.py
+++ b/pytmc/bin/stcmd.py
@@ -154,7 +154,7 @@ def main(tsproj_project, *, name=None, prefix=None,
         [
             jinja2.PackageLoader("pytmc", "templates"),
             jinja2.FileSystemLoader(template_path),
-         ]
+        ]
     )
     jinja_env = jinja2.Environment(
         loader=jinja_loader,
--- a/pytmc/bin/stcmd.py
+++ b/pytmc/bin/stcmd.py
@@ -45,6 +45,12 @@ def build_arg_parser(parser=None):
     )
 
     parser.add_argument(
+        '--hashbang', type=str,
+        default='../../bin/rhel7-x86_64/adsIoc',
+        help='Indicates to the shell which binary to use for the st.cmd script'
+    )
+
+    parser.add_argument(
         '--binary', type=str, dest='binary_name',
         default='adsIoc',
         help='IOC application binary name'
@@ -149,7 +155,8 @@ def jinja_filters(**user_config):
 def main(tsproj_project, *, name=None, prefix=None,
          template_filename='stcmd_default.cmd', plc_name=None, dbd=None,
          db_path='.', only_motor=False, binary_name='ads', delim=':',
-         template_path='.', debug=False, allow_errors=False):
+         template_path='.', debug=False, allow_errors=False,
+         hashbang='../../bin/rhel7-x86_64/adsIoc'):
     jinja_loader = jinja2.ChoiceLoader(
         [
             jinja2.PackageLoader("pytmc", "templates"),
@@ -220,6 +227,7 @@ def main(tsproj_project, *, name=None, prefix=None,
                                'is not an issue.')
 
     template_args = dict(
+        hashbang=hashbang,
         binary_name=binary_name,
         name=name,
         prefix=prefix,
--- a/pytmc/bin/iocboot.py
+++ b/pytmc/bin/iocboot.py
@@ -11,10 +11,9 @@ import pathlib
 
 import jinja2
 
-from . import db, util
-from .. import pragmas
+from . import util
 
-from ..parser import parse, Symbol, separate_by_classname
+from ..parser import parse
 
 
 DESCRIPTION = __doc__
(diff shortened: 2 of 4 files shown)


pytmc/bin/stcmd.py
```diff
@@ -139,9 +144,15 @@ def pragma(eval_ctx, obj, key, default=''):
 def main(tsproj_project, *, name=None, prefix=None,
          template_filename='stcmd_default.cmd', plc_name=None, dbd=None,
          db_path='.', only_motor=False, binary_name='ads', delim=':',
-         debug=False):
+         template_path='.', debug=False):
+    jinja_loader = jinja2.ChoiceLoader(
```
teddyrendahl: I believe this is duplicated from #91?

klauer: Oops - that'll disappear after rebase - sorry.

ZLLentz:
I'm going to run this with my new project and see what happens

ZLLentz:
My test run looks pretty good! (I thought it was completely broken, but my pragmas were malformed). I had forgotten about the `_RBV` convention so some of my PVs look silly, but other than that I only had a few issues. Note that I'm testing with this PR, with #90, with https://github.com/pcdshub/ads-ioc/pull/7, and with https://github.com/pcdshub/ads-ioc/pull/8. I'll ping around to those after writing something up here.

1. For `pytmc stcmd`, it's not intuitive how I need to give a relative path to my `.tsproj` but an absolute path to ads-ioc, though it retrospect it seems fine
2. Some things are wrong with the Makefile, I'll paste my build output in the appropriate PR
3. Some of my PVs dropped their prefix, e.g. should have been `TST:PPM:SFM:RAW_RBV` but instead was just `RAW_RBV`.
4. `st.cmd` did not notice that I had motors and omitted that section

I'm going to debug some of this after bouncing around to the other PRs I was testing.

klauer:
@ZLLentz Thanks for giving it a go! Maybe we can come up with a way of warning the user in instances where it resembles a malformed pragma...

I'll need more details (i.e., the project itself) to debug the other issues you've discovered. If you or @teddyrendahl have some bandwidth/interest in pursuing it, by all means go for it. I've got my work cut out for me this week on the other coast. :)

ZLLentz:
@klauer the intention was for Teddy and I to debug this, though we may send questions your way with precise examples. I expect we'll be able to resolve most of everything. I want my PPM IOC working this week.

ZLLentz:
@n-wbrown  and I talked about the duplication issue and figured it out. There are three sections in a tmc file with symbols, `Inputs`, `Outputs`, and `Internal`. Any input or output variable gets an entry as a `Symbol` in the corresponding section with the full address of the variable as the name, e.g. `Main.struct.variable`. This is redundant with the base `Main.struct` which also gets a `Symbol` entry in the tmc file in the non-io section. Since we iterate through the sub-items of the struct, we hit the io variables twice, with the standalone io symbol being the bogus one that should be ignored, since its pragma does not have any parent PV to build on.

We think the practical (kind of ugly?) solution is to skip entries that have more than one dot in the name, e.g. `Main.fbPPM` is a valid symbol, `GVL.bPump` is a valid symbol, `Main.fbPPM.stGige.bPower` is not because it's either guaranteed to be a duplicate, or if not a duplicate it does not correspond with any top-level pragma in `Main`.

ZLLentz:
The missing PVs in my project were simply due to a typo in the pragma, I put `PV` instead of `pv`

klauer:
Great debugging work, @ZLLentz!

ZLLentz:
I've kludged the culling of invalid symbols and switching the default back to io, and it worked.

I've tracked down an issue where my PVs never get identified as waveforms, when the `SingularChain` is initialized all information about the array gets dropped for these `SubItem` instances. I'm not going to follow that rabbit hole any deeper because I'm running low on time to get my IOC out the door, I'm going to manually edit the generated files...

I'll also defer the auto-generation of st.cmd fixes, I just want to get the PPM to the point where we can do screens for it properly for assembly checkout.

klauer:
Sorry to hear that @ZLLentz. Looks like we still have a lot of work left here. 😦

klauer:
(Rebased and force-pushed to pick up AMS net ID fixes.)

teddyrendahl:

PR #93, opened 2019-07-26
ENH: support concept of "extras" in stcmd generation

klauer:
Support additional template arguments from the command-line, either through directly specified `VAR=VALUE`, from a json string, or from a json file.

A thing I noticed about keeping only one syntax for the `VAR=VALUE`, but allowing it to be specified multiple times. This would force user-specified extras to always be lists instead of scalar values. As a workaround, I added the alternate syntax of `:=` which makes it scalar.

Current functionality overrides automatically generated arguments. This may or may not be a bad thing - perhaps these settings should be relegated into `template_args['argparse_args'] = {'VAR': 'VALUE'}` or something similar.

I'm open to suggestions about any/all of the above.

### DIFF
--- a/pytmc/bin/stcmd.py
+++ b/pytmc/bin/stcmd.py
@@ -6,7 +6,9 @@ Relies on the existence (and linking) of FB_MotionStage function blocks.
 """
 
 import argparse
+import collections
 import getpass
+import json
 import logging
 import pathlib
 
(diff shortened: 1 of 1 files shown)


ZLLentz:
Concept seems good, but I'm confused by the CLI syntax. My understanding of the code as-written is that I do the following if I want a list:
`--extra my_list=some --extra my_list=text` -> `['some', 'text']`
Or the following if I want a list of one element:
`--extra my_elem=ent` -> `['ent']`
Or the following if I want a single value:
`--extra my_value:=worth` -> `'worth'`

... But of course, these all have to be strings? Since we're coming from CLI and substituting into a st.cmd file anyway, do we need lists at all? Can't we just have a super simple `--extra my_text=something` interface?

If we do need lists, why two separate operators (where one is hard to use due to repeated invocation of `--extra` to get a list) instead of something like:
`--extra my_list=[some,text]`
`--extra my_elem=[ent]`
`--extra my_value=worth`

klauer:
I think 
```
--extra my_list=[some,text]
--extra my_elem=[ent]
--extra my_value=worth
```
is exactly what this should have been - thanks, @ZLLentz! And perhaps renamed `macro` just for consistency with other tools. I'll change this up today.

klauer:
For the record: I still haven't gotten around to redoing this - this PR should not be merged.

PR #103, opened 2019-08-13
FIX: default io type should be io, as per the docs

ZLLentz:
https://slaclab.github.io/pytmc/pragma_usage.html#io

### DIFF
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -171,7 +171,7 @@ class TwincatTypeRecordPackage(RecordPackage):
         ValueError
             If unable to determine IO direction
         """
-        io = self.chain.config.get('io', 'i')
+        io = self.chain.config.get('io', 'io')
         if 'o' in io:
             return 'output'
         return 'input'


teddyrendahl:

klauer:
... on second thought, let's discuss this...

ZLLentz:
The alternate PR would be one that changes the docs appropriately.

teddyrendahl:
I'd vote for having the default not be "writable". I think the consequences of accidentally exposing an internal field is greater than not having a field be writable. 

That being said I'm not completely against @ZLLentz suggestion of removing the ambiguity and raising an error (or at least a warning if no `io` is specified)

ZLLentz:
I think this is an important enough choice that I'd want it to have no default. If there was a default, I don't really care which we choose, I just want the docs and the code to be consistent. I wrote the ppm lib stuff assuming the io default, got burned, and I only want to rewrite it once.

klauer:
@n-wbrown interesting - a prototyping mode could be useful. Thoughts @ZLLentz @teddyrendahl?

ZLLentz:
I don't think adding an io line significantly slows down your prototyping speed. I wouldn't oppose these sorts of modes but I don't expect to get much use out of them personally.

n-wbrown:
Merging to bring code behavior and docs in line with one another. We can certainly revisit and change this at a later date if anyone develops strong feelings for the topic.

PR #104, opened 2019-08-13
FIX: Ignore invalid pytmc symbols

ZLLentz:
This gets rid of two kinds of invalid symbols:
- duplicates, where symbol leafs are doubled-up with their containing
  structure
- invalid symbols, where there is not a complete pragma path from
  top-level down to the symbol itself

The tmc file can have duplicate or bogus pytmc symbol entries for anything that is marked as an input or an output variable. The correct symbols to use are the symbols with only two dot-separated fields, such as `GVL.bInput` or `Main.stStruct`, not `Main.stStruct.bOutput` which is actually a double-count.


teddyrendahl:

PR #110, opened 2019-08-14
ENH: error on duplicate records generated

### LINKED ISSUE #96: Warn/error when duplicate PVs are generated from a .TMC file
Currently, this requires validation with a dbd via `pytmc db --dbd path/to/dbd path/to/tmc`. This should be caught regardless of if a dbd file is specified.

What happens when loading the database into the IOC:
* If not of the same type, the IOC will throw an error when loading the database
* If of the same type, this may fail silently on the IOC, causing confusion

Ref #95

klauer:
Closes #96 

Looks like:
```sh
$ pytmc db TC_project/TwinCAT\ Project1/TwinCAT\ Project1/Untitled1/Untitled1.tmc
ERROR:pytmc.bin.db:Duplicate records encountered:
    ADS_TST:ReadOnly:Str (2)
    ADS_TST:ReadOnly:bArray (2)
    ADS_TST:ReadOnly:bState (2)
```

### DIFF
--- a/pytmc/bin/db.py
+++ b/pytmc/bin/db.py
@@ -105,6 +105,19 @@ def process(tmc, *, dbd_file=None, allow_errors=False,
                for record in record_packages_from_symbol(symbol)
                ]
 
+    record_names = [record.pvname for record in records]
+    if len(record_names) != len(set(record_names)):
+        dupes = {name: record_names.count(name)
+                 for name in record_names
+                 if record_names.count(name) > 1
+                 }
+        message = '\n'.join(['Duplicate records encountered:'] +
+                            [f'    {dupe} ({count})'
+                             for dupe, count in sorted(dupes.items())])
+        logger.error(message)
+        if not allow_errors:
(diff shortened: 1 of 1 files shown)


pytmc/bin/db.py
```diff
@@ -105,6 +105,19 @@ def _show_context_from_line(rendered, from_line):
...
+    record_names = [record.pvname for record in records]
+    if len(record_names) != len(set(record_names)):
+        dupes = {name: record_names.count(name)
+                 for name in record_names
+                 if record_names.count(name) > 1
+                 }
+        message = '\n'.join(['Duplicate records encountered:'] +
+                            [f'    {dupe} ({count})'
+                             for dupe, count in sorted(dupes.items())])
+        logger.error(message)
```
teddyrendahl: If you raise the error do you also need to log?

klauer: Perhaps not... 🤷‍♂

PR #111, opened 2019-08-16
FIX: code parsing failure

klauer:


klauer:
I think my parsing code is just too naïve, as somewhat expected. It's either patch this up and make it "good enough" or eventually fall back to tried-and-true (but complicated) grammar-based parsing.

teddyrendahl:
Good enough for me! Just curious if this was intended to catch operator errors or not.

PR #118, opened 2019-09-09
FIX: pragmas on motionstage

klauer:
There are some finer points about this that need to be discussed at some point, but for now this PR just fixes the usage of a pytmc pragma on an FB_MotionStage in conjunction with the stcmd generation script:

Bug:
```
Traceback (most recent call last):
  ...
  File "/reg/g/pcds/epics/R7.0.2-2.0/ioc/ads-ioc/v0.0.2/iocBoot/templates/st.cmd.template", line 63, in top-level template code
    epicsEnvSet("MOTOR_PREFIX",    "{{motor|epics_prefix}}")
  File "/usr/local/lib/python3.7/site-packages/pytmc/bin/stcmd.py", line 137, in epics_prefix
    return get_name(obj, user_config)[0]
  File "/usr/local/lib/python3.7/site-packages/pytmc/bin/stcmd.py", line 118, in get_name
    config = pragmas.squash_configs(*configs)
  File "/usr/local/lib/python3.7/site-packages/pytmc/pragmas.py", line 190, in squash_configs
    squashed['pv'].append(config.pop('pv', None))
AttributeError: 'tuple' object has no attribute 'pop'
```

Looks like `pragmas` functions were tweaked and `stcmd` was not updated to follow that - another hole in the test suite.

Now an `FB_MotionStage` with a pragma of `pv: my:prefix:AXIS1` will correctly generate the following:
```
epicsEnvSet("MOTOR_PREFIX",    "my:prefix:")
epicsEnvSet("MOTOR_NAME",      "AXIS1")
```
and a filter of the form `{{motor|pragma('precision', 3) }}` will get the `precision:` field from the `FB_MotionStage` pragma as well.

PR #120, opened 2019-09-14
ENH: switch to using DUT_MotionStage

### LINKED ISSUE #98: Consider switching to DUT_MotionStation instead the function block FB_MotionStage
Some relevant discussion in https://github.com/pcdshub/ads-ioc/pull/7

ZLLentz:
closes #98 
based on #118, so merge that one first

DUT_MotionStage is the interface we're connecting to in the IOC, not FB_MotionStage. This switches some of the symbol stuff to the data type, which simplifies the handling a bit.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
-[submodule "tests/projects/vonhamos-motion"]
-	path = tests/projects/vonhamos-motion
-	url = ../../pcdshub/vonhamos-motion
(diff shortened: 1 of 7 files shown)


ZLLentz:
rebased on top of new master

ZLLentz:
This is as far as I intend to take it for now. I fixed the tests I broke and added a new one that does some basic parsing of st.cmd output. I think this tests is best kept minimal because I want it to remain relevant as we tweak small details in the template.

PR #122, opened 2019-09-18
FIX: Use precise link name

ZLLentz:
Using the search for two different strings method fails when the axis
count exceeds 10, since searching for M1 also collides with M10, M11,
M12, etc.

We couldn't use the precise name when we were targeting FB_MotionStage,
but now we can because DUT_MotionStage has a clearly defined namespace.

### DIFF
--- a/pytmc/parser.py
+++ b/pytmc/parser.py
@@ -882,16 +882,15 @@ class Symbol_DUT_MotionStage(Symbol):
 
         That is, how the NC axis is connected to the DUT_MotionStage
         '''
-        expected = ('^' + self.name.lower(), '.nctoplc')
+        expected = '^' + self.name.lower() + '.axis.nctoplc'
         links = [link
                  for link in self.plc.find(Link, recurse=False)
-                 if all(s in link.a[1].lower() for s in expected)
+                 if expected in link.a[1].lower()
                  ]
 
         if not links:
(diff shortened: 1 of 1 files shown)


pytmc/parser.py
```diff
@@ -882,16 +882,15 @@ def nc_to_plc_link(self):
 
         That is, how the NC axis is connected to the DUT_MotionStage
         '''
-        expected = ('^' + self.name.lower(), '.nctoplc')
+        expected = '^' + self.name.lower() + '.axis.nctoplc'
         links = [link
                  for link in self.plc.find(Link, recurse=False)
-                 if all(s in link.a[1].lower() for s in expected)
+                 if expected in link.a[1].lower()
```
klauer: Maybe:
```suggestion
                 if expected == link.a[1].lower()
```

ZLLentz: I'll try this really quick to make sure I used the right string

ZLLentz: This actually broke it, I need to figure out what is wrong with the string I used

ZLLentz: `PlcTask Inputs^Main.M1.Axis.NcToPlc` is the full string. Unfortunately this code block doesn't have information about what the task name is supposed to be. I think we just leave it for now.

klauer: Works for me!

PR #123, opened 2019-10-25
ENH/FIX: allow solution file to be passed in for `pytmc summary`

klauer:
Moving the project-from-sln regex in from ads-deploy-docker.

### DIFF
--- a/pytmc/parser.py
+++ b/pytmc/parser.py
@@ -5,6 +5,7 @@ import collections
 import logging
 import os
 import pathlib
+import re
 import types
 
 import lxml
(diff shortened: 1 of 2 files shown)


pytmc/parser.py
```diff
@@ -36,6 +41,33 @@ def parse(fn, *, parent=None):
...
+    '''
+    Find project filenames from a solution.
+
+    Parameters
+    ----------
+    fn : str, pathlib.Path
+        Solution filename
+    exclude : list or None
+        Exclude certain extensions. Defaults to excluding .tcmproj
+    '''
```
n-wbrown: +1 Docs

PR #128, opened 2019-11-04
FIX: declaration/implementation text can be None

klauer:
Closes pcdshub/ads-deploy#6

Declaration/implementation sections in source code files is allowed to be `None`. This fixes `pytmc summary` to not bail in such cases.

I'm sure we'll find similar issues like these as time goes on and more projects are run through the parser. Unless the issue reports get out of hand, I'd rather fix them one-by-one instead of handling such exceptions generically.


klauer:
Locally confirmed this works with https://github.com/slacAWallace/pmps-dev-arbiter/commit/4bdda0c09e2dd7469f991374c248bf4916656fa0, as it was where @slacawallace found the bug.

PR #129, opened 2019-11-06
FIX: enums; utf8 BOM on win32

klauer:
The UTF-8 BOM was causing pytmc running on Windows to bail immediately with the message:
```
lxml.etree.XMLSyntaxError: Start tag expected, '<' not found, line 1, column 1
```

Enums were apparently also being ignored in some (all?) cases in recent builds. b79f4a9 attempts to fix that by ensuring `DataType.walk()` is always yielding something, allowing for the creation of chains.

Confusing implementation detail: enums should be seen to pytmc in a way similar to a built-in type (UINT + some metadata), and as `BuiltInType` yields `[]` for `walk()` this PR brings it in line with that.

~One test is now failing as a previously-ignored enum is now found to be duplicated...~ The pragma was already fixed upstream; updated to master in 53aa27f.

### DIFF
--- a/pytmc/parser.py
+++ b/pytmc/parser.py
@@ -34,7 +34,7 @@ def parse(fn, *, parent=None):
     '''
     fn = case_insensitive_path(fn)
 
-    with open(fn, 'rt', encoding='utf-8') as f:
+    with open(fn, 'rb') as f:
         tree = lxml.etree.parse(f)
 
     root = tree.getroot()
--- a/pytmc/parser.py
+++ b/pytmc/parser.py
@@ -673,11 +673,8 @@ class DataType(_TmcItem):
 
         if hasattr(self, 'SubItem'):
             for subitem in self.SubItem:
-                if condition and not condition(subitem):
-                    continue
-                for item in subitem.walk():
-                    if condition is None or all(condition(it) for it in item):
-                        yield [subitem] + item
+                for item in subitem.walk(condition=condition):
+                    yield [subitem] + item
 
     @property
     def enum_dict(self):
@@ -737,7 +734,8 @@ class SubItem(_TmcItem):
         return f'{namespace}.{type_.text}' if namespace else type_.text
 
     def walk(self, condition=None):
-        yield from self.data_type.walk(condition=condition)
+        if condition is None or condition(self):
+            yield from self.data_type.walk(condition=condition)
 
 
 class Module(_TmcItem):
@@ -878,8 +876,7 @@ class Symbol(_TmcItem):
     def walk(self, condition=None):
         if condition is None or condition(self):
             for item in self.data_type.walk(condition=condition):
-                if condition is None or all(condition(it) for it in item):
-                    yield [self] + item
+                yield [self] + item
 
     @property
     def array_info(self):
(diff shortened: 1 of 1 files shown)


pytmc/parser.py
```diff
@@ -34,7 +34,7 @@ def parse(fn, *, parent=None):
     '''
     fn = case_insensitive_path(fn)
 
-    with open(fn, 'rt') as f:
+    with open(fn, 'rt', encoding='utf-8') as f:
```
n-wbrown: This isn't make-or-break but something I noticed in the python docs on the the `open` function that might be worth consideration:

> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever locale.getpreferredencoding() returns), but any text encoding supported by Python can be used. See the codecs module for the list of supported encodings.

pytmc/parser.py
```diff
@@ -63,7 +63,9 @@ def projects_from_solution(fn, *, exclude=None):
         for match in SLN_PROJECT_RE.findall(solution_text)
     ]
 
-    return [pathlib.Path(project) for project in projects
+    solution_path = pathlib.Path(fn).parent
+    return [(solution_path / pathlib.Path(project)).absolute()
```
n-wbrown: Is there a case where this syntax prevails over `pathlib.Path(project).absolute()`? I think there's a use case I'm not comprehending right now.

pytmc/parser.py
```diff
@@ -655,6 +657,13 @@ def is_complex_type(self):
         return True
 
     def walk(self, condition=None):
+        if self.is_enum:
+            # Ensure something is yielded for this type - it doesn't
+            # appear possible to have SubItems or use ExtendsType
+            # in this case.
```
n-wbrown: 👍

pytmc/parser.py
```diff
@@ -867,9 +876,10 @@ def info(self):
                     )
 
     def walk(self, condition=None):
-        for item in self.data_type.walk(condition=condition):
-            if condition is None or all(condition(it) for it in item):
-                yield [self] + item
+        if condition is None or condition(self):
```
n-wbrown: Could applying this condition to `SubItem.walk` provide a performance/reliability improvement by skipping the `walk` if that subitem lacks a pragma? I'm a little surprised you haven't added this condition to both methods. Its likely that I'm not fully comprehending how this line helps prevent the enum issue.

pytmc/parser.py
klauer: :+1: good to note! It appears that all Visual Studio-written XML files are UTF-8. 

However, your strange method of "reading the documentation" prompted me to read lxml's for `parse()` - and it turns out it suggests using binary mode and letting it handle encoding.

I'll push another commit with that change.

pytmc/parser.py
klauer: The key here is: `solution_path` points to the `.sln` file somewhere on disk, and `project` is relative to that. `pathlib.Path(project).absolute()` alone is incorrect as it doesn't take into account the solution.

pytmc/parser.py
klauer: I confusingly combined this unrelated fix with this PR, though it isn't really related. I noticed that we weren't checking the condition (which as you note, is checking for a pragma's existence normally).

It should speed things up a bit by first checking that the Symbol has a pragma - an obvious requirement - before building and yielding all of the chains.

pytmc/parser.py
n-wbrown: I see now. Got it!

PR #137, opened 2019-11-25
ENH: add autosave support

### LINKED ISSUE #41: Support Autosave and use sensible defaults
I think it is crucial to allow a user to request that a specific PV created by a `pytmc` database file is archived or autosaved. Ideally this is just an additional pragma, but putting it somewhere where we can make sure that the built IOC respects these takes some thought.

klauer:
Closes #41 

Defaults are specified in classes from `pytmc.record`. These can be overridden by using specific keys in the pragmas:

```
        To apply to either input or output records, pragma keys `autosave_pass0`
        or `autosave_pass1` can be used.

        To only apply to input records, pragma keys `autosave_input_pass0`
        `autosave_input_pass1` can be used.

        To only apply to output records, pragma keys `autosave_output_pass0`
        `autosave_output_pass1` can be used.
```

More info in the documentation commit (see diff).

I have not given the defaults much thought, but with this PR as a basis we can fill in those details later.

### DIFF
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -17,10 +17,11 @@ logger = logging.getLogger(__name__)
 class EPICSRecord:
     """Representation of a single EPICS Record"""
 
-    def __init__(self, pvname, record_type, fields=None, template=None,
-                 autosave=None):
+    def __init__(self, pvname, record_type, direction, fields=None,
+                 template=None, autosave=None):
         self.pvname = pvname
         self.record_type = record_type
+        self.direction = direction
         self.fields = OrderedDict(fields) if fields is not None else {}
         self.template = template or 'asyn_standard_record.jinja2'
         self.autosave = dict(autosave) if autosave else {}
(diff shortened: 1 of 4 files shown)


klauer:
Rebased on master - ready for review.

pytmc/record.py
```diff
@@ -288,6 +332,12 @@ class FloatRecordPackage(TwincatTypeRecordPackage):
     output_rtyp = 'ao'
     dtyp = 'asynFloat64'
     field_defaults = {'PREC': '3'}
+    autosave_defaults = {
+        'input': dict(pass0={'PREC'},
+                      pass1={}),
+        'output': dict(pass0={'VAL', 'PREC'},
+                       pass1={}),
+    }
```
ZLLentz: Am I interpreting these defaults correctly?

for input records, autosave the precision for UI stuff
for output records, autosave the value for configuration stuff and the precision for UI stuff

klauer: Yes, that sounds about right. 

I think I mentioned above somewhere that I didn't give much thought to the defaults here, so I'm very much open to suggestions.

pytmc/record.py
```diff
@@ -33,6 +36,32 @@ def __init__(self, pvname, record_type, fields=None, template=None):
...
+        `autosave_input_pass1` can be used.
+
+        To only apply to output records, pragma keys `autosave_output_pass0`
+        `autosave_output_pass1` can be used.
+
+        Parameters
+        ----------
+        config : dict
+            The pragma configuration dictionary
+        """
```
n-wbrown: 👍

pytmc/record.py
ZLLentz: I don't have any real input, I just wanted to make sure I understood what the code was doing.

Does an autosaved output record on pass 0 get put into the PLC?

klauer: OK, understood. No, pass 0 will not result in a put:

```
Pass 0 indicates restoring information prior to record initialization on IOC
initialization, whereas pass 1 indicates restoring information after record
initialization. Pass 0 is generally safe and does not cause record processing,
whereas pass 1 is just as if one were to ``caput`` to the record after starting
the IOC. When in doubt, use pass 0 and/or ask an EPICS expert.
```
;)

ZLLentz: Ok, so that means if I want to autosave something like a velocity setting that would be pass 1. Do you have an example of a record that we want to be pass 0?

ZLLentz: Ok I guess PREC is a good example of a pass0

klauer:
We'll probably revisit the default autosave fields at some point, but this is probably good-to-go? Will merge by COB unless there are more issues to resolve.

PR #143, opened 2019-11-27
Sort records when rendered

n-wbrown:
Add a new sort method for the EPICSRecord.field Ordered dict. This first sorts all entries specified in a dictionary and then sorts any unidentified items alphabetically. 

As a side not I've tried adding sphinx doc types using the python type hints format. I found this to be a satisfactory


pytmc/record.py
```diff
@@ -29,11 +33,14 @@ def __init__(self, pvname, record_type, fields=None, template=None):
             self.template
         )
 
-    def render(self):
+    def render(self, sort: bool = True):
```
klauer: Maybe required would be better?

pytmc/record.py
```diff
@@ -439,3 +446,59 @@ def generate_output_record(self):
...
+def sort_fields(unsorted: OrderedDict, sort_lookup: Optional[dict] = None,
+                last: Optional[bool] = True) -> OrderedDict:
+    """
+    Sort the ordered dict according to the sort_scheme given at instantiation.
+    Does NOT sort in place.
+
+    Parameters
+    ----------
+
+    unsorted
```
klauer: Don't we still need types after the parameter names in the docstring?

pytmc/record.py
```diff
@@ -439,3 +446,59 @@ def generate_output_record(self):
...
+
+    # Separate items identified by the sort_sceme into instructed_unsorted
+    for x in unsorted:
+        if x in sort_lookup:
+            instructed_unsorted[x] = unsorted[x]
+        else:
+            naive_unsorted[x] = unsorted[x]
+
+    # Separately sort instructed and, naively sorted entries
+    print(instructed_unsorted)
```
klauer: Stray print

pytmc/record.py
n-wbrown: Do you mean requiring the calling function to specify the `sort` argument's value or simply removing the argument and always sorting?

pytmc/default_settings/unified_ordered_field_list.py
```diff
@@ -0,0 +1,605 @@
+# single list for all record types dictating field order
```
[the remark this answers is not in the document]
n-wbrown: My thinking was that I wanted  to avoid overcrowding .py files that we edit more regularly. I can definitely move this to `record.py` though.

pytmc/record.py
n-wbrown: Screen Shot 2019-11-27 at 4 15 45 PM
That was the really cool part about using type hints. The sphinx docs (with the help of sphinx_autodoc_typehints) copies what's specified for function/class so it doesn't need to be written redundantly.

n-wbrown: I'm not quite sure how I feel about the formatting of optional parameters though.

Issue #140, opened 2019-11-26
Pinned global datatypes have a different property name


n-wbrown:
`plcAttribute_pytmc` is inserted into the `Property/Name` for a pinned global data type. 

Our first step is to determine if this is intended Beckhoff behavior. If is, we support the new property name. If it is not an intended we create a temporary fix that can easily be removed if/when beckhoff patches the behavior.

slacAWallace:
Can we prioritize this before finding out if it's intended? Nate is out until next week.

n-wbrown:
I can start working on a patch tomorrow.

n-wbrown:
closed by #144

PR #144, opened 2019-12-04
Pinned Global Variables have different <Name> field 

n-wbrown:
This PR alters the 'pragmas.get_pragma' method to identify pragmas based upon a list of acceptable strings instead of matching against a single string. Default arguments have been updated to match.


pytmc/pragmas.py
```diff
@@ -321,31 +325,58 @@ def find_pytmc_symbols(tmc):
...
+    """
+    if name is None:
+        name = ['pytmc','plcAttribute_pytmc']
+
     if hasattr(item, 'Properties'):
         properties = item.Properties[0]
         for prop in getattr(properties, 'Property', []):
-            if prop.name == name:
+            # Return true if any of the names searched for are found
+            if any([ indiv_name in prop.name for indiv_name in name]):
```
klauer: I think we could drastically simplify all of this, were we to just say that if the user asks for pragma `x` that `plcAttribute_x` is valid as well. It's a unique enough string which we certainly will not use. It would also then allow the pragma `name` arg to be a single, simple string as before.

(Minor thing you should know - pass an iterable to `any` and not a `list`, otherwise you forego the speedup; stop by and I can explain more)

pytmc/pragmas.py
```diff
@@ -321,31 +325,58 @@ def find_pytmc_symbols(tmc):
...
-def get_pragma(item, *, name='pytmc'):
-    'Get all pragmas with a certain tag'
+def get_pragma(item: Union[parser.SubItem, Type[parser.Symbol]], *,
+               name: List[str] = None) -> Generator[str, None, None]:
+    """
+    Get all pragmas with a certain tag.
+
+    Parameters
+    ----------
+    item
```
klauer: We didn't finish up our conversation about this in the last PR - can we keep putting the types in the docstring as well? Otherwise, the rest of the code base will be inconsistent.

pytmc/pragmas.py
```diff
@@ -321,31 +325,58 @@ def find_pytmc_symbols(tmc):
...
+        Representation of beckhoff variable or data structure  
+
+    name
+        Accept tmc entries where the <Name> field equals the passed string 
+    
+    Yields
+    ------
+    str
```
klauer:
```suggestion
    str
        A pragma containing the requested tag
```

tests/conftest.py
```diff
@@ -26,6 +26,22 @@ def dbd_file():
 def tmc_filename(request):
     return request.param
 
+@pytest.fixture(scope='module')
+def tmc_xtes_sxr_plc():
```
klauer: Not used?

tests/conftest.py
```diff
@@ -26,6 +26,22 @@ def dbd_file():
...
+    """
+    generic .tmc file
+    """
+    return TMC_ROOT / "xtes_sxr_plc.tmc"
+
+
+@pytest.fixture(scope='module')
+def tmc_ArbiterPlc():
```
klauer: Think we could add a submodule with the arbiter project at a certain point in time? I rather like that pattern.

tests/conftest.py
n-wbrown: Correct. I just figured that I should be consistent if I was creating one for the other tmc file.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "tests/projects/lcls-twincat-motion"]
 	path = tests/projects/lcls-twincat-motion
 	url = ../../pcdshub/lcls-twincat-motion.git
+[submodule "tests/projects/lcls-twincat-pmps"]
+	path = tests/projects/lcls-twincat-pmps
+	url = https://github.com/pcdshub/lcls-twincat-pmps.git
```
klauer:
```suggestion
	url = ../../pcdshub/lcls-twincat-pmps.git
```

n-wbrown:
@slacAWallace [pmps-dev-arbiter](https://github.com/pcdshub/pmps-dev-arbiter) contains pinned global variables but none of them are instantiated in pytmc. Their top level GVL does not contain a pragma so no pinned global variables are created. To accommodate, I've incorporated the `.tmc` file sent to us on the 26th which *does* include fully instantiated pragmas for testing pytmc's ability to identify and render pinned global variables. What project this `.tmc` file come from?

The pmps-dev-arbiter test is currently marked for `xfail`.

n-wbrown:
closes #140

PR #155, opened 2019-12-09
FIX: sort autosave fields

### LINKED ISSUE #154: Autosave field order not predictable
A diff from my recent rebuild. The only change prior to `make` was swapping `IOC_TOP` to Mike's release instead of his dev build:
```
-  info(autosave_pass0, "VAL PREC")
+  info(autosave_pass0, "PREC VAL")
```

klauer:
Closes #154


ZLLentz:
I'm glad I didn't try to do this because I would have needed to go through 20 pages of jinja2 docs before realizing how to sort

klauer:
It's easy once you've already spent the time looking into those 20 pages of jinja docs ;)

PR #159, opened 2019-12-12
Enh: show incomplete pragmas

### LINKED ISSUE #147: Enhance debug tool to show chains with incomplete pragmas
e.g., a chain of:
`symbol` `struct` `struct.member`
where the `symbol` has a pragma, but nothing in `struct` has a pragma, it would be useful to see that somehow.

### LINKED ISSUE #158: Debug's config chart isn't reliably printing fields
The instead of using the field as the column header, the configuration chart in the debug interface displays the column number and does not display the corresponding configuration

n-wbrown:
Adds a category to the debug interface to show TwinCAT variable chains that have *not* been fully rendered to records. Alterations have been made to the chain generation procedures to make it less dependent upon record generation to allow flawed records to be generated.
closes #147 

Also addresses some oddities in the debug interface's config_info table not drawing properly. 
closes #158

### DIFF
--- a/pytmc/bin/debug.py
+++ b/pytmc/bin/debug.py
@@ -218,14 +218,6 @@ class TmcSummary(QtWidgets.QMainWindow):
                 key = str(key)
                 if key not in columns:
                     columns[key] = max(columns.values()) + 1 if columns else 0
-                print("row:")
-                print(row)
-                print("key:")
-                print(key)
-                print("columns[key]")
-                print(columns[key])
-                print("value:")
-                print(value)
                 # accumulate a list of entries to print in the chart. These
                 # should only be printed after `setColumnCount` has been run
                 if isinstance(value, dict):
@@ -241,8 +233,6 @@ class TmcSummary(QtWidgets.QMainWindow):
 
         items = zip(chain.config['pv'], chain.item_to_config.items())
         for row, (pv, (item, config)) in enumerate(items):
-            print("config:***************")
-            print(config)
             # info_dict is a collection of the non-field pragma lines
             info_dict = dict(pv=pv)
             if config is not None:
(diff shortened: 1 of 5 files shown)


n-wbrown:
Should close #158 for real this time.

pytmc/record.py
```diff
@@ -100,8 +100,14 @@ def __init__(self, ads_port, chain=None, origin=None):
         # Due to a twincat pragma limitation, EPICS macro prefix '$' cannot be
         # used or escaped.  Allow the configuration to specify an alternate
         # character in the pragma, defaulting to '@'.
-        macro_character = self.chain.config.get('macro_character', '@')
-        self.pvname = chain.pvname.replace(macro_character, '$')
+        try:
```
klauer: I think you might want to check if `self.chain.config is None` first, and then just set `pvname = None`

pytmc/bin/debug.py
```diff
@@ -190,42 +204,72 @@ def _update_config_info(self, record):
...
                 if key not in columns:
                     columns[key] = max(columns.values()) + 1 if columns else 0
-                self.config_info.setItem(
-                    row, columns[key], QtWidgets.QTableWidgetItem(str(value))
-                )
+                # accumulate a list of entries to print in the chart. These
+                # should only be printed after `setColumnCount` has been run
                 if isinstance(value, dict):
-                    add_dict_to_table(row, value)
+                    yield from (row, value)
```
klauer: Is `yield from` correct? 
`yield` would result in: `.setItem(row, value)` 
`yield from` would result in: `.setItem(row)` `.setItem(value)`

pytmc/bin/debug.py
```diff
@@ -190,42 +204,72 @@ def _update_config_info(self, record):
...
+                column_write_entries.extend(new_entries)
 
+        # setColumnCount seems to need to proceed the setHorizontalHeaderLabels
+        # in order to prevent QT from incorrectly drawing/labeling the cols
+        self.config_info.setColumnCount(len(columns))
         self.config_info.setHorizontalHeaderLabels(list(columns))
+        # finally print the column's entries
+        for line in column_write_entries:
+            if line:
+                self.config_info.setItem(*line)
```
klauer: Is there a case where `line` can be `None`?

pytmc/record.py
n-wbrown: 👍

pytmc/bin/debug.py
n-wbrown: You're right, this isn't correct. What I should be doing is `yield from _update_confic_info`.

pytmc/bin/debug.py
n-wbrown: No. This was added in response to an intermediate bug that I have since fixed.

PR #161, opened 2019-12-13
CI: switch to pcds-ci-helpers for linting

klauer:


klauer:
A minor CI change - self-merging.

PR #162, opened 2019-12-13
ENH: add preliminary archive file support

### LINKED ISSUE #139: Support the EPICS archiver appliance
From Murali, we now know that:

```
Cron jobs are in place running processRequestFile.py, so just make sure your IOC build and
release process installs archive request files to $IOC_DATA/$IOC_NAME/archive/*.archive w/ syntax
MY:PV:NAME    period    [ type ]
period can be integer or floating point
type is optional, either monitor or scan, defaults to scan.
```

We can create our own `info` tags per-record and generate these files. Or `pytmc` could output them directly.

klauer:
Closes #139

See the related issue + docs for further information.

- [x] Fix nested imports
- [x] Add better tests
- [x] Docs

### DIFF
--- a/pytmc/bin/db.py
+++ b/pytmc/bin/db.py
@@ -6,12 +6,14 @@ m-epics-twincat-ads driver.
 
 import argparse
 import logging
+import os
 import sys
 
 from collections import defaultdict
 
 from .. import linter, parser
 from ..pragmas import find_pytmc_symbols, record_packages_from_symbol
+from ..record import generate_archive_settings
 
 
 logger = logging.getLogger(__name__)
@@ -119,10 +121,11 @@ def process(tmc, *, dbd_file=None, allow_errors=False,
     if exceptions and not allow_errors:
         raise LinterError('Failed to create database')
 
-    if allow_no_pragma:
-        record_names = [record.pvname for record in records if record.valid]
-    else:
-        record_names = [record.pvname for record in records]
+    record_names = [single_record.pvname
+                    for record_package in records if record_package.valid
+                    for single_record in record_package.records
+                    ]
+
     if len(record_names) != len(set(record_names)):
         dupes = {name: record_names.count(name)
                  for name in record_names
@@ -167,17 +170,6 @@ def build_arg_parser(parser=None):
     parser.formatter_class = argparse.RawTextHelpFormatter
 
     parser.add_argument(
-        'tmc_file', metavar="INPUT", type=str,
-        help='Path to interpreted .tmc file'
-    )
-
-    parser.add_argument(
-        'record_file', metavar="OUTPUT", type=str,
-        nargs='?',
-        help='Path to output .db file'
-    )
-
-    parser.add_argument(
         '--dbd',
         '-d',
         default=None,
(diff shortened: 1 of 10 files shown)


klauer:
As you'll likely be one of the first users, @ZLLentz, feel free to chime in if the PR makes sense from the perspective of defaults + pragmas. Review-wise @n-wbrown has things covered - unless you feel so inclined.

pytmc/pragmas.py
```diff
@@ -45,6 +45,16 @@
                        2)      # 2.0 Hz - 2 updates every 1 second
 
 
+_ARCHIVE_RE = re.compile(
+    # Rate portion (e.g., 1 s, 1s, 1Hz, 1 Hz)
+    r'^(?P<rate>\d*\.\d+|\d+)\s*(?P<hz_or_sec>hz|s)'
```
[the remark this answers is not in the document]
klauer: Note the `re.IGNORECASE` below in the flags 😉

pytmc/bin/db.py
```diff
@@ -201,11 +193,52 @@ def build_arg_parser(parser=None):
...
+        help=('Do not write the archive file, regardless of OUTPUT '
+              'filename settings.')
+    )
+
+    parser.add_argument(
+        'tmc_file', metavar="INPUT", type=str,
+        help='Path to interpreted .tmc file'
+    )
+
+    class OutputFileAction(argparse.Action):
```
klauer: FYI, this is some argparse magic that handles a few scenarios:
1. User passes in `tmc_file` but no `output_file.db` - result is that the db is written to standard output
2. User passes in `tmc_file` and specifies `output_file.db` - result is that the database is written to `output_file.db` and the archive file is written to `output_file.db.archive`
3. User passes in `tmc_file` and specifies `output_file.db` and `--no-archive-file`, results in only the database being written to `output_file.db`

pytmc/bin/db.py
```diff
@@ -201,11 +193,52 @@ def build_arg_parser(parser=None):
         help='Do not show db file context around errors'
     )
 
+    archive_group = parser.add_mutually_exclusive_group()
+    archive_group.add_argument(
+        '--archive-file',
+        type=argparse.FileType('wt', encoding='ascii'),
```
klauer: I learned this argparse type from @hhslepicka - it takes care of opening the file with these options, passing you the file object instead of the string. Pretty handy, so I used it in a couple spots here.

It tends to make our other functions such as `main` here more reusable, as well. Instead of requiring a filename, we can pass it a file-like object that works however we desire and it will work just the same.

pytmc/bin/pytmc.py
```diff
@@ -56,6 +58,13 @@ def main():
         formatter_class=argparse.RawTextHelpFormatter
     )
 
+    top_parser.add_argument(
+        '--version', '-V',
+        action='version',
```
klauer: This is a new one as well that we forgot (or I may have inadvertently nuked in a refactor, sorry!). The action helper simplifies things such that we don't require the `version` kwarg in our `main` function below.

pytmc/record.py
```diff
@@ -194,7 +209,7 @@ def from_chain(*args, chain, **kwargs):
             spec = StringRecordPackage
         else:
             try:
-                spec = data_types[data_type.name]
+                spec = DATA_TYPES[data_type.name]
```
klauer: This is an unrelated change, but I think it should be noted that globals should be all uppercase, where sensible. 

FWIW, `logger` tends to be an exception for us that we live with.

n-wbrown:
This all looks very solid to me. You can merge when ready.

PR #169, opened 2020-01-23
ENH/FIX: Pointers

ZLLentz:
- Add `is_pointer` property to `Symbol` tmc items to check if they are pointers or not
- Add aux info and expand `pytmc summary` info to include pointer/array text
- filter motors down to skip pointers for the stcmd generation

closes #168

PR #170, opened 2020-01-31
FIX: off by one error in complex array unrolling

ZLLentz:
The off-by-one error in array bounds fix in #169 created a new off-by-one error in the code introduced in #121 that was relying on the previously off-by-one value. This fixes the issue.


ZLLentz:
Yeah, I suppose it didn't really matter, but now it matches the numbers you designate in the beckhoff PLC.  I checked that nothing else relied on this already but I was basically just hoping that the tests would pass :man_shrugging:

PR #183, opened 2020-02-22
BUG: pass arguemnt to db.process

### LINKED ISSUE #182: st.cmd used with --allow-errors does not pass this argument onto db.process
`pytmc stcmd` fails due to `raise LinterError('Failed to create database')` when the `--allow-errors` flag is used even when `pytmc db --allow-errors` successfully creates a database. It appears that `db.process` is not being given an argument for the `allow_errors` parameter when called from `pytmc.bin.stcmd.main`.

n-wbrown:
Resolve #182


klauer:
Looks good to me 👍 - just fix the flake8 issue and I'll merge.

PR #184, opened 2020-02-24
Describe process error

n-wbrown:
Provide the name of TwinCAT variable when the db.process fails to build a record from the chain. 

Example output:
```
ERROR:pytmc.bin.db:Error creating record: GVL_GMD.fb_EM1K0_GMD_GPI_10.PG.eState 'EnumInfo' object has no attribute 'Enum'
```
Open to feedback as always.

### DIFF
--- a/pytmc/bin/db.py
+++ b/pytmc/bin/db.py
@@ -116,7 +116,7 @@ def process(tmc, *, dbd_file=None, allow_errors=False,
                   if isinstance(ex, Exception)]
 
     for ex in exceptions:
-        logger.error('Error creating record: %s ', ex)
+        logger.error('Error creating record: %s', ex)
         records.remove(ex)
 
     if exceptions and not allow_errors:
--- a/pytmc/bin/db.py
+++ b/pytmc/bin/db.py
@@ -113,10 +113,10 @@ def process(tmc, *, dbd_file=None, allow_errors=False,
     ]
 
     exceptions = [ex for ex in records
-                  if isinstance(ex, dict)]
+                  if isinstance(ex, Exception)]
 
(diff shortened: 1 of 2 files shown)


pytmc/pragmas.py
```diff
@@ -586,6 +586,6 @@ def record_packages_from_symbol(symbol, *, pragma: str = 'pytmc',
             yield RecordPackage.from_chain(symbol.module.ads_port, chain=chain)
         except Exception as ex:
             if yield_exceptions:
-                yield ex
+                yield {'ex': ex, 'tcname': chain.tcname}
```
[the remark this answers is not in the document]
n-wbrown: 👍

pytmc/pragmas.py
```diff
@@ -586,6 +586,6 @@ def record_packages_from_symbol(symbol, *, pragma: str = 'pytmc',
             yield RecordPackage.from_chain(symbol.module.ads_port, chain=chain)
         except Exception as ex:
             if yield_exceptions:
-                yield ex
+                yield type(ex)(f"{chain.tcname} {ex.args[0]}")
```
klauer: If you re-create the exception, you lose the stack trace info.

pytmc/bin/db.py
```diff
@@ -116,7 +116,7 @@ def _show_context_from_line(rendered, from_line):
                   if isinstance(ex, Exception)]
 
     for ex in exceptions:
-        logger.error('Error creating record: %s', ex)
+        logger.error('Error creating record: %s ', ex)
```
klauer:
```suggestion
        logger.error('Error creating record: %s', ex)
```

PR #192, opened 2020-03-24
MNT: require the user to provide ZNAM and ONAM for BinaryRecordPackage

### LINKED ISSUE #190: BOOL ONAM/ZNAM Defaults
Currently the defaults for `BinaryRecordPackage` are:  https://github.com/slaclab/pytmc/blob/7290ba67a91fca2970a4311cf3a165ab7c5cc696/pytmc/record.py#L471

This is unintuitive because the only TwinCAT type that maps to `BinaryRecordPackage` is `BOOL`: https://github.com/slaclab/pytmc/blob/7290ba67a91fca2970a4311cf3a165ab7c5cc696/pytmc/record.py#L637

Since `BOOL` is the only TwinCAT type that uses `BinaryRecordPackage` and because `BOOL` shows up as `TRUE` or `FALSE` while online in TwinCAT, the default should be changed to `TRUE` and `FALSE`, since the usage is overwhelmingly as a boolean flag rather than as a number. In fact, in our projects there isn't a single case where the default values are used intentionally.

If there is some disagreement here then I will soon go through all of my TwinCAT code and copy/paste the same field lines to all of my `BOOL` pragmas.

ZLLentz:
closes #190 
alternative to #191

### DIFF
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -119,7 +119,7 @@ class RecordPackage:
     the necessary setup before the record can be configured.
     """
     _required_keys = {}
-    _required_fields = {}
+    _required_fields = []
     archive_fields = []
 
     def __init__(self, ads_port, chain=None, origin=None):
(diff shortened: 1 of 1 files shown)


klauer:
Closed in favor of #191

PR #224, opened 2020-08-07
FIX: link fallback to project if saved incorrectly

klauer:
Let's see if we can fix the project first...


klauer:
OK, for future reference: 'save PLC as independent file; save; rebuild' was sufficient here.

PR #226, opened 2020-08-18
DOC: release v2.7.1 + include release notes in documentation

klauer:
Took a bit longer than anticipated - had to start from scratch here.


klauer:
Thanks for taking a look - no problem 👍

PR #236, opened 2020-11-12
FIX: fix autosave pass1 by setting PINI=YES

### LINKED ISSUE #235: pass1 Autosave is Broken
And I think `pytmc` is the right place to fix it.
Nobody has been able to prove this issue to me, so I did some tests today.

I created a PLC program with the following pragmas:
```
    // Floating point setting, default autosave
    {attribute 'pytmc' := '
        pv: PLC:TST:PASS0
    '}
    fAutoPass0: LREAL;

    // Floating point setting, pass1 autosave
    {attribute 'pytmc' := '
        pv: PLC:TST:PASS1
        autosave_pass1: VAL
    '}
    fAutoPass1: LREAL;
```

Then I did some caputs to get to this state:
```
pass0 = 3.1415
pass1 = 42
```

Wait a bit (I did some laundry), restart the PLC to zero them out, restart the IOC to activate the autosave load:
```
pass0 = 0
pass1 = 0
```

... "What?" I thought. So there's an actual issue. I see the successful loading messages in the ioc shell but the values aren't here in the PLC. I expected the pass1 to make it through and the pass0 to fail to get into the PLC, but both have failed.

For a prospective fix, I decided to set PINI=1 by manually editing the db files. I did this on both the pass0 and the pass1 PVs.
Then, I put 21 to each of the values on the PLC side so I could be sure if "something" was happening (in retrospect this step did not need to be done).

I restarted the IOC with the PINI=1 hotfix on my IOC and found:
```
pass0=21
pass1=42
```
... so without the PINI set, the record never gets processed, even if it is in pass1. And with the PINI set, the pass0 autosave does not get applied to the PLC.

My proposed solution options:
- Set PINI=1 in all records
- OR, more conservatively, set PINI=1 in all records with VAL autosaved at pass1

Thoughts?

ZLLentz:
closes #235 

This is a workaround fix for autosave pass1 not behaving as expected for records where PINI=0

Implementation:
- Add a method `update_pini_for_autosave` on `EPICSRecord` with an appropriate docstring
- Call this method when creating output records

With the goal of being able to cleanly remove this method if we figure out why the expected behavior did not work at ads-ioc in https://github.com/pcdshub/ads-ioc/issues/65

Resulting diff on my test IOC (extended to show the full record):
```
$ git diff
diff --git a/iocBoot/ioc-tc-mot-example/tc_mot_example.db b/iocBoot/ioc-tc-mot-example/tc_mot_example.db
index 319a356..cd3628b 100644
--- a/iocBoot/ioc-tc-mot-example/tc_mot_example.db
+++ b/iocBoot/ioc-tc-mot-example/tc_mot_example.db
@@ -13,6 +13,7 @@ record(ai, "PLC:TST:PASS1_RBV") {

 record(ao, "PLC:TST:PASS1") {
   field(DESC, "Main.fAutoPass1")
+  field(PINI, "1")
   field(DTYP, "asynFloat64")
   field(OUT, "@asyn($(PORT),0,1)ADSPORT=851/Main.fAutoPass1=")
   field(PREC, "3")
   info(autosaveFields, "VAL")
   info(autosaveFields_pass0, "PREC")
   info(archive, "VAL")
 }
```

### DIFF
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -125,6 +125,19 @@ class EPICSRecord:
                     fields = set(config[config_key].split(' '))
                     self.autosave[record_key] = fields
 
+    def update_pini_for_autosave(self):
+        """
+        Set PINI=1 if VAL is autosaved at pass1.
+
+        This is a bandaid fix for unexplained behavior where the
+        autosave pass1 does not send values to the PLC unless
+        PINI=1.
+
+        This is intended to be called for output records.
+        """
+        if 'VAL' in self.autosave['pass1']:
+            self.fields['PINI'] = 1
+
     def render(self, sort: bool = True):
         """Render the provided template"""
         for field, value in list(self.fields.items()):
@@ -509,6 +522,7 @@ class TwincatTypeRecordPackage(RecordPackage):
         _update_description(record, self.chain.tcname)
 
         record.update_autosave_from_pragma(self.config)
+        record.update_pini_for_autosave()
         return record
 
     @property


klauer (approved):
This looks good. I'd like a bit more time to research this before rolling out the bandaid, but would not complain if it were deemed necessary to.

ZLLentz:
I'd like to deploy the bandaid during the upcoming PAMM next week if we can't find the real culprit. This would help a lot with the tmo and fee vacuum systems.

klauer:
Sounds like a plan. Will make a goal today to dive a bit deeper, ensuring there's enough time to tag/rebuild/release IOCs for the PAMM.

klauer:
I'm convinced this is the correct fix, not a band-aid. Let's move forward with it.

PR #243, opened 2020-11-20
ENH: subitem pragma support

### LINKED ISSUE #223: Allow for partial array pragmas

### LINKED ISSUE #242: Debug tool reports missing GUIDs when loading a TMC file
Issue
------
* Debug tool reports missing GUIDs when loading a TMC file
* Database generation works as expected

Example
---------

`ERROR:pytmc.bin.db:Error creating record: Symbol GVL_Logger.fbRootLogger failure: Type with GUID {18071995-0000-0000-0000-000000000018} not in TMC`

Details
--------
This appears to be getting hit only in the debug tool as opposed to the database generation as the debug tool is exhaustive: it looks at _all_ chains, including those with pragmas missing somewhere.

In looking through these chains, it finds GUIDs that are not defined in the TMC, but rather those found in the TwinCAT installation files (for things like `PVOID`, `GUID`, etc.)

klauer:
Closes #223 
Fixes as a follow-up to #240 
Closes #242 

Feature point: extends pragma syntax, allowing for forward-references to sub-items (e.g., member variables of function blocks).

If `fbTest` contains a member variable `variable1` with a different `io` setting than desirable for the particular use case, this PR now allows for changing it by way of adding this pragma:

```
variable1.io: io
```

with the overall syntax of `[member].[sub-member].pragma_key: pragma_value`.

Started off with a slightly modified `kfe-motion.tmc` file ([here](https://gist.github.com/klauer/ec6bb07179fafdccc9cacd04f56476a9)), setting `arrStates.array: 1..4` "externally".

Appears to be working locally, but I suspect I won't be catching all edge cases. Going to re-run pytmc on some IOCs and compare the output now.

Debug tool does not show these pragmas cleanly. Not sure I'll have the time to fix it this go around.

### DIFF
--- a/docs/pragma_usage.rst
+++ b/docs/pragma_usage.rst
@@ -468,3 +468,58 @@ autosaving:
       ...
       info(autosaveFields_pass0, "VAL DESC")
    }
+
+
+Sub-Items
+.........
+
+There is a special syntax to pragma a member of a structure differently
+depending on the instance.
+
+Given this example function block:
+
+.. code-block:: none
+
+    FUNCTION_BLOCK FB_Test
+    VAR
+        {attribute 'pytmc' := 'pv: variable1'}
+        variable1 : LREAL := 0.0;
+
+        {attribute 'pytmc' := 'pv: variable2'}
+        variable2 : LREAL := 0.0;
+    END_VAR
+    END_FUNCTION_BLOCK
+
+Adding these pragmas to two instances:
+
+.. code-block:: none
+
+    {attribute 'pytmc' := '
+        pv: TEST:A
+        variable1.io: i
+        variable2.io: i
+    '}
+    fbTestA : FB_Test;
+
+    {attribute 'pytmc' := '
+        pv: TEST:B
+        variable1.io: io
+        variable2.io: io
+    '}
+    fbTestB : FB_Test;
+
+The above would result in two different ``io`` settings for
+``fbTestA.variable1`` and ``fbTestB.variable1``.
+
+There are some limitations on this preliminary feature. You may not add or
+change the following fields for now:
+
+.. code-block:: none
+
+    pv
+    link
+    field
+
+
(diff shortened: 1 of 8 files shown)


klauer:
I didn't see anything strange when re-running `make db` on all the PLC projects I have locally. Based on the tests passing and those results, I'm at least reasonably certain I didn't screw something up majorly here.

ZLLentz:
Is there any part of this you want a close review on? I'm having trouble following all the codepaths just reading through. I'll definitely give this a try with the cclib states things (since that is a patch that needs to happen anyway).

klauer:
Yeah, it's admittedly a confusing mess, despite starting off kinda sorta clean - the usual story.

Testing it out as a user is good enough as a reviewer here, I think: verifying that it doesn't screw up your db for one, and seeing if it fits enough of what you might want in the future to consider #223 closed.

ZLLentz:
Got sidetracked-
The diff looks very promising:
https://github.com/ZLLentz/lcls-plc-kfe-motion/commit/56581ac5a4da11857955f4cadad414b2459b76c4

A little puzzled by the culling of my st.cmd though- I haven't changed anything here.

klauer:
Oof, your test saved me (us) a lot of headaches - I neglected to double-check `st.cmd`. I think I know where to look for the cause; will fix it up and push shortly.

ZLLentz (approved):
Ran it again- output looks great

Adding an item to my todo list to apply this feature to the common component plcs :+1:

klauer:
Thanks, @ZLLentz! Here goes nothing...

PR #260, opened 2021-04-23
MNT: fix remaining slaclab references

### LINKED ISSUE #259: Docs URL on PiPy is pointing to slaclab and is broken

klauer:
Closes #259 

Will require new release to fix PyPI, but in principle this should do it.
Done using `git sed` (*) so should double check I didn't do anything terribly wrong.

(*) 
`.gitconfig`
```
[alias]
sed=!git grep -z --full-name -l '.' | xargs -0 sed -i '' -e
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ version: ~> 1.0
 
 env:
   global:
-    # Doctr deploy key for slaclab/pytmc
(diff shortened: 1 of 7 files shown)


ZLLentz (approved):
lgtm

PR #262, opened 2021-04-27
ENH: add scale and offset support for ints and floats

### LINKED ISSUE #261: Pragma enhancement: "scale" key
Request
---------

Ability to scale integer (or floating point) values from the PLC on the EPICS side in a consistent way.

"scale" key
-----------

e.g., 
```
scale: 1.234
offset: 10.0
```

Presence of a `scale` key on a record normally generated as `longin` would then create an `ai` record instead.

Alternative: `calc` key
--------------------

e.g., `calc: 2 * value + 1.54`

Maybe consider a `calc` field which may allow generic calculations by way of an auto-generated `calc` record.
This could then allow for easy slope, offset, and _perhaps_ even references to other symbols. (That last part is iffy, though)

cc @slacAWallace

klauer:
Closes #261 

### Details (borrowed from docs in PR)

Integer and floating point values may have an EPICS-side scaling applied.

Example:

    scale: 3.0
    offset: 1.0

Values will be scaled according to the following:

    readback_value = raw_value * scale + offset
    setpoint_value = (user_value - offset) / scale

Note:

    If either ``scale`` or ``offset`` are applied to an integer symbol, the
    generated EPICS record type will no longer be a "long" integer input/output
    record but rather change to an analog input/output record.

    Keep this in mind if using advanced "field" directives in your pragmas.

If unspecified, ``scale`` will be assumed to be 1.0 and ``offset`` 0.0.

### DIFF
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -306,6 +306,12 @@ class RecordPackage:
                     f'Unsupported data type {data_type.name} in chain: '
                     f'{chain.tcname} record: {chain.pvname}'
                 ) from None
+            if spec is IntegerRecordPackage:
+                if "scale" in chain.config or "offset" in chain.config:
+                    # longin/longout do not support scaling. Special-case and
+                    # cast to a FloatRecordPackage.
+                    spec = FloatRecordPackage
+
         return spec(*args, chain=chain, **kwargs)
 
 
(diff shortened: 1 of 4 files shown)


pytmc/record.py
```diff
@@ -572,6 +578,37 @@ class FloatRecordPackage(TwincatTypeRecordPackage):
...
+        scale = "1.0" if scale is None else scale
+        offset = offset or "0.0"
+        return scale, offset
+
+    def generate_input_record(self):
+        record = super().generate_input_record()
+        scale, offset = self.get_scale_offset()
+        if scale is not None:
+            # If LINR==SLOPE, VAL = RVAL * ESLO + EOFF
+            record.fields["LINR"] = "SLOPE"
```
ZLLentz, line 597: I didn't realize this was just built in, that's neat

ZLLentz (approved):
Sorry I got distracted halfway through this review- looks good. Everything seems well-motivated and doing it using the ai/ao built-ins is great. Good addition to the docs as well.

ZLLentz:
Yeah, I was planning to do pcds-envs this week- though you might understand literally everything is stalled for the moment.

klauer:
Oh, and I didn't mention it originally, but this style of "tell me what you want to happen - scale/offset" versus "specifically give me an analog input record" is the direction I hope future pragmas additions will continue in. It'll offer more flexibility in the end.

Yeah, the network outage has been terribly disruptive for almost everything I've been trying to do. I can only imagine how much more it has affected your tasks...

PR #272, opened 2021-11-10
MNT: miscellaneous minor fixes and tweaks

klauer:
## Changes
* Some things sitting around uncommitted for a while
* Some additional type hint cleanups/fixes
* One (probably never-hit) bug fixed: "lines_between" had a logic error that would not search for the start marker if the end marker was found before it
* Reduced memory consumption slightly by not caching the xml element on every TwincatItem
* Add `EnumerationTextList` with `get_source_code` support; as previously these translatable types were missing

## Parser classes background
So there are some unfortunate things in the way that the parser breaks out `TwincatItem` classes. Here's the background on why the parser items are as they are:
* I wanted easy mappings of TwinCAT `<xml-tag>` to Python class, where methods and properties could then be context-specific (e.g., _give me the source code of the GVL_)
* I wanted things like `gvl.Declaration` or `plc.TcSmItem.PlcProject.Mappings` to be a thing you could do, since I thought `__getitem__`-based `plc["TcSmItem"]["PlcProject"]["Mappings"]` was verbose and ugly
* I was unaware of which XML items would be able to have more than one entry in a given scope. (In retrospect, TwinCAT has xsds that provide this information)
* This led to the current (ugly) structure of `plc.TcSmItem[0].PlcProject[0].Mappings[...]`, along with all of the code that grabs the only element from an assumed single-element list

## What about better type hints?
Now, type hints weren't _really_ a thing when pytmc was getting on its feet, at least with the versions we used and whatnot.
So type hints in `TwincatItem` subclasses are unfortunately really bad, given how the classes are structured:
```python
class EnumInfo(_TmcItem):
    # Some have type hints like this now
    Comment: list  
    # What we really want to say to describe the structure as-is
    Comment: Optional[typing.List[Comment]]  
    # With a refactor to de-listify single items, this would be ideal
    Comment: Optional[Comment]  
```
The latter two appear to be circular references rather than one to a corresponding `Comment` class.

Where do we go from here? Though I'd really like to see this improved, I guess we just leave it as-is until a real refactor could be justified - unless anyone has a better idea.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
 
 jobs:
   allow_failures:
-    - name: "Python 3.8 - PIP"
+    - name: "Python - PIP"
 
 after_failure:
   - cat logs/run_tests_log.txt
--- a/pytmc/bin/stcmd.py
+++ b/pytmc/bin/stcmd.py
@@ -228,7 +228,7 @@ def main(tsproj_project, *, name=None, prefix=None,
                                f'Projects: {list(by_name)}')
 
     symbols = separate_by_classname(plc.find(Symbol))
-    motors = [mot for mot in symbols['Symbol_DUT_MotionStage']
+    motors = [mot for mot in symbols.get('Symbol_DUT_MotionStage', [])
               if not mot.is_pointer]
 
     if plc.tmc is None:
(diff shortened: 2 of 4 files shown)


ZLLentz (approved):
Everything here seems sensible to me :+1:, just cleanups and bugfixes. Nothing terribly large or new to nitpick.

Issue #278, opened 2022-02-28
Document PV linking functionality


klauer:
In use in several places but apparently not documented on the pragma page.

Example usage and suggested usage on the PLC end:
https://github.com/pcdshub/lcls-twincat-general/blob/18895a44d33d2f9c465841e08f2a6a2169be56dc/LCLSGeneral/LCLSGeneral/POUs/Data/FB_LREALFromEPICS.TcPOU#L5-L20

PR #279, opened 2022-03-01
TST: does linking work as expected?

klauer:
After opening #278 this reminded me of a local test branch to see if `link` was doing what I thought it should (or if I remembered how I had even designed it, perhaps).

It turns out that it does, fortunately.

Should add some docs to this PR prior to merging.


ZLLentz (approved):
good tests are good

PR #283, opened 2022-05-20
FIX: CP link instead of CPP link

### LINKED ISSUE #282: pytmc link settings should be "CP" instead of "CPP"
Code:
https://github.com/pcdshub/pytmc/blob/bd2529195298e5b1dbfd96772a7702b2199f4f4b/pytmc/record.py#L496

Otherwise, we will be stuck at the 2Hz updates in the record:

> If the input link specifies CA, CP, or CPP, regardless of the location of the process variable being referenced, it will be forced to be a Channel Access link. This is helpful for separating process chains that are not tightly related. If the input link specifies CP, it also causes the record containing the input link to process whenever a monitor is posted, no matter what the record’s SCAN field specifies. If the input link specifies CPP, it causes the record to be processed if and only if the record with the CPP link has a SCAN field set to Passive. In other words, CP and CPP cause the record containing the link to be processed with the process variable that they reference changes.

klauer:
## Context

Closes #282 

Prior to this PR, we are erroneously limited to 2Hz updates for all IOC->PLC PV links due to the 0.5 second SCAN rate configured for linked records. What we want is the IOC to write to the PLC on *every* update, and failing that at 2Hz.

## Background

* A `CPP` link means "channel access link which causes processing (when the monitor is fired), **only if** the record's scan field is set to Process Passive
* A `CP` link means "channel access link which causes processing (when the monitor is fired), regardless of the record's SCAN field

In our case, we want the linked record to cause processing on *every* update. This is how the mechanism is supposed to work with pytmc writing values from the IOC to the PLC.

### DIFF
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -214,7 +214,7 @@ class RecordPackage:
         return config
 
     def _configure_link(self):
-        'Link this record to a pre-existing EPICS record via a CA (CPP) link'
+        'Link this record to a pre-existing EPICS record via a CA (CP) link'
         self.linked_to_pv = self.config.get('link') or None
 
     def _configure_pvname(self):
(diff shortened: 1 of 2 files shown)


ZLLentz (approved):
CI failure looks like the pyqt thing and the new pre-commit requirements
If this passes tests locally and makes the IOC behave as you expect I think we can merge

ZLLentz:
Code change + test fixes looks good- merge at will if it was passing locally
Should follow-up later to fix the CI troubles

PR #287, opened 2022-06-24
ENH: autosave field additions

### LINKED ISSUE #286: Add fields to autosave defaults
* Alarm limit and severity fields should be autosaved for all records
* Description fields probably should also be autosaved

Anything else?

klauer:
## Context

Closes #286 

## Changes
* Updates the vendored `ads.dbd` to the ads-ioc R0.5.0 database definition file (this makes the diff line count look massive...)
    * This is used in the test suite to verify the fields that we have set
* Adds description field to autosave for all records, input and output
* Adds alarm severity and limit fields to autosave for all relevant input and output records
* Adds control limit (drive low/high) fields to autosave for relevant output records

## Helpful whatrecord blurb

```python
import whatrecord
dbd = whatrecord.parse("pytmc/tests/ads.dbd")
for rtyp in ("ai", "ao", "bi", "bo", "mbbi", "mbbo", "longin", "longout", "waveform"):
    print(f"\n\n{rtyp}\n{'-' * len(rtyp)}")
    for name, fld in dbd.record_types[rtyp].fields.items():
        if fld.prompt and "sev" in fld.prompt.lower():
            print(f'"{name}",  # {fld.prompt}')
```

### DIFF
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -3,7 +3,7 @@ Record generation and templating
 """
 import logging
 from collections import ChainMap, OrderedDict
-from typing import Optional
+from typing import Dict, List, Optional, Set
 
 import jinja2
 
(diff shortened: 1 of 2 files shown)


ZLLentz (approved):
I looked at this with the intention of reviewing the choices for defaults which all look correct to me. I did a double-take because UDFS seems to be missing from the EPICS record reference manual.

I think there are some people who would want the enum state names to be autosaved but I think it's correct for that not to be a default.

klauer:
> I think there are some people who would want the enum state names to be autosaved but I think it's correct for that not to be a default.

Right - my thought was that should be opt-in as well.

PR #293, opened 2022-08-10
REF: sort generated records by TwinCAT symbol name (tcname)

### LINKED ISSUE #292: EPICS database record sorting options (or the default?)
Databases are generated from the order that pytmc discovers symbols inside the tmc file.

Some things:
* I had thought that the order was more stable than it is.
* Parts of it are often stable but other parts jiggle around depending on project rebuilds
* pytmc not having to find all symbols and sort them was considered a minor speedup

I'd vote to change this to either:
* Order records by the full twincat symbol name. (I think this could be more stable than PV name.)
* Order records by their name

klauer:
Closes #292 

There's 3 lines of implementation; the remainder is clean-up/annotation:

```python
    def by_tcname(record: RecordPackage):
        return record.tcname

    records.sort(key=by_tcname)
```

I was wrong about the iterative generation of the records as in #292. Finding chains is an iterative process but generating `db` files from the pared down record list is not. So, there's no real performance impact here.

### DIFF
--- a/pytmc/bin/db.py
+++ b/pytmc/bin/db.py
@@ -10,10 +10,11 @@ import os
 import pathlib
 import sys
 from collections import defaultdict
+from typing import List, Optional, Tuple
 
 from .. import linter, parser
 from ..pragmas import find_pytmc_symbols, record_packages_from_symbol
-from ..record import generate_archive_settings
+from ..record import RecordPackage, generate_archive_settings
 
 logger = logging.getLogger(__name__)
 DESCRIPTION = __doc__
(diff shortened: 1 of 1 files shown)


pytmc/bin/db.py
```diff
@@ -121,10 +146,20 @@ def _show_context_from_line(rendered, from_line):
...
+    def by_tcname(record: RecordPackage):
+        return record.tcname
+
+    records.sort(key=by_tcname)
+
+    record_names = [
+        single_record.pvname
+        for record_package in records
+        if record_package.valid
+        for single_record in record_package.records
```
ZLLentz, line 161: This multi list comprehension totally breaks my brain because the first clause is a variable in the second for, rather than in the first, because it's ordered like a nested for loop instead of as a "sentence".

I didn't even realize you could multi-for in a list comprehension.

ZLLentz (approved):
Looks good to me, should stabilize the git diffs on rebuild

pytmc/bin/db.py
klauer: Yeah, I don't much like it. If I rewrote this, it'd be in two steps or with a generator.

PR #302, opened 2023-04-04
DEV/MNT: migrate to GitHub Actions and pyproject.toml [LCLSPC-603]

klauer:
## Description
* Upgrading by way of https://github.com/pcdshub/pcds-python-migration-tools
* pytmc has been migrated to use setuptools-scm, replacing versioneer, as its version-string management tool of choice.
* pytmc has been migrated to use the modern ``pyproject.toml``, replacing ``setup.py`` and related files.
* Older language features and syntax found in the repository have been updated to Python 3.9+ standards by way of ``pyupgrade``.
* Sphinx 6.0 is now supported for documentation building.
* ``docs-versions-menu`` is now used for documentation deployment on GitHub Actions.  
* Removed the no-longer-required ``doctr`` documentation deployment key.
* `pyupgrade` was used to update annotations, though there might not have been many here
* `pycln` was used to clean up imports
* Using the same Jira ticket number so we can pretend it was done as part of the same effort, when in fact we forgot about poor old pytmc
* Moved lcls-twincat-motion in test suite to latest master
* Reformatted codebase with ``black``

## Motivation and Context
* Migrating because we want to migrate our projects from Travis CI.

## How Has This Been Tested?
* GitHub Actions

## Where Has This Been Documented?
This PR and Jira [LCLSPC-603](https://jira.slac.stanford.edu/browse/LCLSPC-603)


docs/release_notes.rst
```diff
@@ -2,6 +2,21 @@
  Release History
 =================
 
+v2.12.0 (2023-04-??)
```
ZLLentz, line 5: I think you should jam a number in there and we'll tag after
```suggestion
v2.12.0 (2023-04-04)
```

conda-recipe/meta.yaml
```diff
@@ -1,50 +1,54 @@
...
 extra:
   recipe-maintainers:
-    - klauer
-    - hhslepicka
-    - zllentz
-    - n-wbrown
+  - klauer
+  - hhslepicka
+  - zllentz
+  - n-wbrown
```
tangkong, line 54: Should we update this now that Nolan and Hugo aren't with SLAC anymore?  I suppose this should happen on conda-forge primarily.

tangkong (approved):
Looks pretty standard here.  This one's quite the doozy though

ZLLentz (approved):
Looks good, just the one suggestion

conda-recipe/meta.yaml
klauer: Good call - removed the section entirely as it's not used here.

Also noticed at the same time that `coverage` is an unnecessary dev requirement and removed it.

docs/release_notes.rst
klauer: Done 👍

PR #308, opened 2023-06-07
ENH: handle ST_MotionStage/DUT_MotionStage rename and aliasing

ZLLentz:
In https://github.com/pcdshub/lcls-twincat-motion/pull/163, one of the things I do is rename `DUT_`-prefixed structs to `ST_` to be in-line with naming conventions. I leave the `DUT_` version behind as a deprecated alias of the `ST_` version, so as far as PLC projects go this is pretty transparent: leaving the code as-is gives a warning but continues to work, update the name to get rid of the warning.

This caused a problem here in `pytmc` where we ascribe specific meaning to `DUT_MotionStage`. This is actually the second time we've had such a problem.

This PR renames `DUT_MotionStage` support to `ST_MotionStage` and adds a mechanism for symbols to define a list of aliases, adding `DUT_MotionStage` as an alias of `ST_MotionStage`.

This did help my open demo/example PR include all the previously missing motors.
See [here](https://github.com/pcdshub/lcls-plc-example-motion/blob/422f6f8fa6c472843da4833b5f4516ac1a267b72/lcls-plc-example-motion/tc_mot_example/POUs/Main.TcPOU) for the source code where I have a mix of both symbol names and [here](https://github.com/pcdshub/lcls-plc-example-motion/pull/10/commits/422f6f8fa6c472843da4833b5f4516ac1a267b72) for the difference after using this diff.

I've added and adjusted unit tests as appropriate, including adding a real motion lib project for completeness.

In addition, I made a small tweak to the version file. I'm using PYTHONPATH munging with some symbolic link magic to test this and I noticed that the git version checker doesn't resolve the link. This caused an issue because the `ads-ioc` `Makefile` checks the `pytmc` version.


pytmc/version.py
```diff
@@ -28,13 +28,14 @@ def __init__(self):
...
-        repo_root = Path(__file__).resolve().parent.parent
+        here = Path(__file__).resolve()
+        repo_root = here.parent.parent
         if (repo_root / ".git").exists() or (repo_root / ".git_archival.txt").exists():
             try:
                 # Git checkout
                 from setuptools_scm import get_version
 
-                return get_version(root="..", relative_to=__file__)
+                return get_version(root="..", relative_to=here)
```
klauer, line 38: Should we bring this fix to our project template and update our other projects when we get a chance?

klauer (approved):
Fixes look good to me, and the migration to `ST_MotionStage` is fine to me. The extra alias machinery may come in handy in the future.

pytmc/version.py
```diff
@@ -28,13 +28,14 @@ def __init__(self):
 
     def _get_version(self) -> Optional[str]:
         # Checking for directory is faster than failing out of get_version
-        repo_root = Path(__file__).resolve().parent.parent
+        here = Path(__file__).resolve()
+        repo_root = here.parent.parent
```
tangkong, line 32: These changes seem like they should be equivalent, do you have any idea why your symbolic link magic made this necessary?

tangkong (approved):
Changes make sense to me 👍

tangkong (approved):
Changes make sense to me 👍

pytmc/parser.py
```diff
@@ -1551,8 +1558,10 @@ def summary_type_name(self):
         return self.data_type.summary_type_name
 
 
-class Symbol_DUT_MotionStage(Symbol):
```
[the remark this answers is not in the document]
ZLLentz: Is it worth throwing in a
```
Symbol_DUT_MotionStage = Symbol_ST_MotionStage
```
just in case?

pytmc/version.py
ZLLentz: I think so, I should PR to the template prior to merging this one so it isn't forgotten.
Getting a null version is only a problem here because pytmc's version is actively checked, I don't think this causes issues in general.
Alternatively, I can change my habits to stop my PYTHONPATH rampage and use venv to test these sorts of things instead.

pytmc/version.py
ZLLentz: it's the get_version line that was the main functional fix:
```
# __file__ does not resolve symbolic links
return get_version(root="..", relative_to=__file__)
```
vs
```
# here resolves symbolic links via resolve()
return get_version(root="..", relative_to=here)
```

pytmc/version.py
ZLLentz: Made a PR in the cookiecutter repo

pytmc/parser.py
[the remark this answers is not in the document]
klauer: It couldn't hurt, I think

PR #310, opened 2023-06-13
FIX: macro sub in st.cmd motor records

ZLLentz:
Previously, the `@` macro substitution machinery did not run for the st.cmd motor records.
This PR removes the duplicated logic and paths us through the same record generation code used for general-purpose db files. This fixes the aforementioned issue.

See https://github.com/pcdshub/lcls-plc-example-motion/pull/10/commits/15bc045b52321a7968d50101f573b8d4a7148177 for mild proof that this works. I've conflated the fix here with some unrelated edits to make my pvnames shorter (they were too long) and edits I made to add parentheses in my substitution lines, but you can see the `@` symbol turning into a `$`

I've also added a unit test.

### DIFF
--- a/pytmc/bin/stcmd.py
+++ b/pytmc/bin/stcmd.py
@@ -12,7 +12,7 @@ import pathlib
 
 import jinja2
 
-from .. import pragmas
+from .. import pragmas, record
 from ..parser import NC, Symbol, parse, separate_by_classname
 from . import db, util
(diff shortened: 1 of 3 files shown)


klauer (approved):
This looks right to me 👍

tangkong (approved):
Looks good to me. ✅

Issue #299, opened 2022-11-09
`pytmc template` filename usage awkward on Windows


klauer:
`os.pathsep` is currently used to delimit the input file and the output file:

https://github.com/pcdshub/pytmc/blob/20d35b3c6328ae1e5ced222b82f05f79dd3ee4ee/pytmc/bin/template.py#L714

This is `:` on Linux and `;` on Windows.
Our ads-ioc Makefile only uses `:` (https://github.com/pcdshub/ads-ioc/blob/fc8a48a586751be00ba62b79c3a6cac7aface529/iocBoot/templates/Makefile.base#L101)

This is used with ads-deploy in Visual Studio, resulting in something like:
```
	--template "c:/Repos/ads-ioc/R0.6.0//iocBoot/templates/st.cmd.template":"st.cmd" \
```

My thought as to a fix:

I think we should standardize on `:` for all platforms. 
On Windows, we can iteratively check for the longest match that is an existing file. The remainder is the output.
It'll be easier for us and users in the end... probably?

For now, this is patched locally/temporarily on plcprog-console

PR #317, opened 2023-07-27
FIX: hotfix for couple issues found on plcprog-console

klauer:
* Related to #299 but may not necessarily close it (it's a bit of a hack)
* Patches over bizarre `DUT_MotionStage` aliasing issue where sphinx was failing to find the `nc_axis` attribute of `Symbol_ST_MotionStage` instances and trying to use `__getitem__` on it (`x.nc_axis` -> `x["nc_axis"]`) and throwing nonsensical errors in the meantime
  * This was evaluated from this line in the IOC startup script: https://github.com/pcdshub/ads-ioc/blob/98b6bbd49c2433ae9ddc962ec229732cfa8ecbde/iocBoot/templates/st.cmd.template#L156C17-L156C17

### DIFF
--- a/pytmc/bin/template.py
+++ b/pytmc/bin/template.py
@@ -751,7 +751,7 @@ def main(
 
     all_rendered = {}
     for template in templates:
-        input_filename, output_filename = template.split(os.pathsep, 1)
+        input_filename, output_filename = template.rsplit(":", 1)
         if input_filename == "-":
             # Check if it's an interactive user to warn them what we're doing:
             is_tty = os.isatty(sys.stdin.fileno())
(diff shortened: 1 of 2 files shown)


ZLLentz:
This did pass the unit tests last time, right? I can't comprehend how these changes would cause the st.cmds to generate without the motion sections...

klauer:
So perhaps I fixed `DUT_MotionStage` and broke `ST_MotionStage` with the parsing utilities?
More to investigate tomorrow :sigh: ...

ZLLentz:
This behavior is reproducible offline using my random lfe motion plc project clone

ZLLentz:
This must be related to the specifics of the template that I edited to search for ST_MotionStage

klauer:
Back to draft with this one - not an appropriate fix here. 

This quick "fix" hid the issue rather than resolved it. Hmm...

klauer:
Yeah, upon further investigation this is very much not the right fix. Test suites to the rescue!

PR #320, opened 2023-07-28
REF: always use colon delimiter for `pytmc template`

klauer:
Closes https://github.com/pcdshub/pytmc/issues/299

For `pytmc template`, always use a colon delimiter, regardless of the platform.
That means that the input file should be of the form: `input_filename:output_filename` on Windows, Mac, Linux.
(Previously this could have been `input_filename;output_filename` on Windows)

Some special handling is then required to parse these appropriately, since Windows paths can have colons in them. Two absolute paths could then look like `C:\a\b.c.txt:C:\d\e\f.txt` - confusingly enough. The algorithm here checks to see if the first file exists in the case where the delimiter itself is ambiguous.

A quicker hack/fix is already out on plcprog-console: (`filename.rsplit(":", 1)`) - because we only use one absolute filename, the output is relative to the IOC directory. 

I don't think this is the cleanest solution, but it is _a_ solution...

### DIFF
--- a/pytmc/bin/template.py
+++ b/pytmc/bin/template.py
@@ -751,7 +751,7 @@ def main(
 
     all_rendered = {}
     for template in templates:
-        input_filename, output_filename = template.split(os.pathsep, 1)
+        input_filename, output_filename = template.rsplit(":", 1)
         if input_filename == "-":
             # Check if it's an interactive user to warn them what we're doing:
             is_tty = os.isatty(sys.stdin.fileno())
--- a/pytmc/bin/template.py
+++ b/pytmc/bin/template.py
@@ -715,6 +715,53 @@ def _split_macro(macro: str) -> tuple[str, str]:
     return tuple(parts)
 
 
+def split_input_output(arg: str) -> tuple[str, str]:
+    """
+    Split user-specified template argument into input/output file.
+
+    Parameters
+    ----------
+    arg : str
+        The user-specified command-line argument for ``--template``.
+
+    Returns
+    -------
+    input_filename : str
+        The input filename.
+    output_filename : str
+        The output filename.
+    """
+    delim = ":"
+    num_delim = arg.count(delim)
+    if num_delim == 0:
+        raise ValueError(
+            f"Invalid input: {arg!r} should be in the format "
+            f"'input_filename:output_filename'"
+        )
+
+    if num_delim == 1:
+        # Our job is easy in this scenario
+        return tuple(arg.split(delim, 1))
+
+    input_parts = arg.split(delim)
+    if input_parts[0] == "-":
+        # Special case: - as input file means read /dev/stdin
+        return "-", delim.join(input_parts[1:])
+
+    output_parts = []
+    while input_parts:
+        input_filename = delim.join(input_parts)
+        if os.path.exists(input_filename):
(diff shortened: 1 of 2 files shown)


pytmc/bin/template.py
ZLLentz, line 1: Quick Q: what behavior is this fixing? What happens on windows if we use the semicolon path sep?

klauer: I believe (as in https://github.com/pcdshub/pytmc/issues/299) it was that ads-ioc Makefile was executing pytmc with a colon delimiter for input/output but pytmc (on Windows) was expecting a semicolon.

So the makefile would have to be written to be platform-dependent and I didn't want to do that, instead opting for tweaking how pytmc handled the input.

pytmc/bin/template.py
```diff
@@ -715,6 +715,53 @@ def _split_macro(macro: str) -> tuple[str, str]:
...
+    input_parts = arg.split(delim)
+    if input_parts[0] == "-":
+        # Special case: - as input file means read /dev/stdin
+        return "-", delim.join(input_parts[1:])
+
+    output_parts = []
+    while input_parts:
+        input_filename = delim.join(input_parts)
+        if os.path.exists(input_filename):
+            return input_filename, delim.join(output_parts)
```
tangkong, line 755: What about the incredibly antagonistic case where we have `C:/a/b/c:D:/a/b/c` and the file `C` exists in the current directory? 

Could we check that the joined output parts also exists as a file? I think in cases like this the output_parts won't resolve

pytmc/bin/template.py
ZLLentz: Makes 100% sense

ZLLentz (approved):
Implementation looks good, should always work with all well-formed paths on linux and windows

pytmc/bin/template.py
tangkong: Another absurd edge case might be if the entire path exists as is, leaving output_parts empty?

ZLLentz: Can we do something smart here but assuming that path parts are either devoid of colons, or have a single-letter drive name and one colon?

klauer: I think the important part is "does the input file exist so we can load it as a template" and then "try to write the output file that's specified after that"

To that end I think this does its job well enough. That said I'd be happy to implement another algorithm but I'm not sure if any would be able to capture all the potential weird edge cases...

tangkong: I guess I'd be satisfied if we could somehow verify the output path was valid, so in the `C:/a/b/c:D:/a/b/c` -> `(C, /a/b/c:D:/a/b/c)` we could see that `/a/b/c:D:/a/b/c` might raise.  But I suppose output files can be arbitrarily whack.  

I think I agree, this probably does the job well enough

tangkong (approved):

pytmc/bin/template.py
klauer: > verify the output path was valid

I _think_ the best way there is to try and open it and see what happens. You could see if the parent directory exists and is writeable, the filename exists and has the correct permissions - but in the end it's up to `open(..., "wt")` to have the final word.

klauer:
(Merging so we can test this with Patrick's updated project)

PR #321, opened 2023-07-31
MNT: back-compat for 'pytmc template' delimiter-less filename

klauer:
(Short follow-up PR to #320 )

``pytmc template``, which takes a TwinCAT project and jinja template source
  code to generate project-specific output, now expects all platforms to use
  the same delimiter (``":"``) for template filename patterns. Examples include:

  * Read template from ``a.template`` and write expanded version to ``a.txt``:
    ``pytmc template my.tsproj --template a.template:a.txt``
  * Read template from ``a.template`` and write results to standard output:
    ``pytmc template my.tsproj --template a.template``       **This PR fixes this scenario for back-compat purposes**
  * Read template from standard input and write results to standard output:
    ``pytmc template my.tsproj --template -``
  * Read template from standard input and write results to ``/path/to/a.txt``:
    ``pytmc template my.tsproj --template -:/path/to/a.txt``

### DIFF
--- a/pytmc/bin/template.py
+++ b/pytmc/bin/template.py
@@ -729,15 +729,15 @@ def split_input_output(arg: str) -> tuple[str, str]:
     input_filename : str
         The input filename.
     output_filename : str
-        The output filename.
+        The output filename.  Empty filename implies stdout.
     """
     delim = ":"
+
     num_delim = arg.count(delim)
     if num_delim == 0:
-        raise ValueError(
-            f"Invalid input: {arg!r} should be in the format "
-            f"'input_filename:output_filename'"
-        )
+        # Without a delimiter, assume the file is the template filename
+        # and the user wants to write to standard output
+        return arg, ""
 
     if num_delim == 1:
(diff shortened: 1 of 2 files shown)


tangkong (approved):
After quickly checking how this function was used, LGTM.  👍

pytmc/bin/template.py
```diff
@@ -729,15 +729,15 @@ def split_input_output(arg: str) -> tuple[str, str]:
...
+
     num_delim = arg.count(delim)
     if num_delim == 0:
-        raise ValueError(
-            f"Invalid input: {arg!r} should be in the format "
-            f"'input_filename:output_filename'"
-        )
+        # Without a delimiter, assume the file is the template filename
+        # and the user wants to write to standard output
+        return arg, ""
```
ZLLentz, line 740: This should always be safe to assume because later we check that the input filename exists, and sending to stdout is never going to cause a problem

ZLLentz (approved):
This change looks good to me

PR #328, opened 2024-07-19
DEV: Making _RBV PVs read-only

KaushikMalapati:
https://jira.slac.stanford.edu/browse/ECS-5005

This adds an ASG field with the value NO_WRITE to any EPICSRecord with direction input in the jinja database template. This access security group will be added to a new version of ads-ioc and will only allow reading, not writing to pvs. 

https://github.com/pcdshub/ioc-common-ads-ioc/pull/103

### DIFF
--- a/pytmc/templates/asyn_standard_record.jinja2
+++ b/pytmc/templates/asyn_standard_record.jinja2
@@ -22,4 +22,7 @@ record({{record.record_type}}, "{{record.pvname}}") {
   info(archive, "{{ record.archive_settings['method'] }} {{ record.archive_settings['seconds'] }}: {{ record.archive_settings['fields'] | join(' ') }}")
   {% endif %}
 {% endif %}
+{% if record.direction == "input" %}
+  field(ASG, "NO_WRITE")
+{% endif %}
 }
--- a/pytmc/tests/test_record.py
+++ b/pytmc/tests/test_record.py
@@ -54,11 +54,10 @@ def test_output_record_with_write_access():
         "record_type": "ao",
         "direction": "output",
     }
-
     ec = EPICSRecord(**kwargs)
     record = ec.render()
     assert "ASG" not in record
-    
+
 
 def test_sort_fields():
     unsorted_entry = OrderedDict(
--- a/pytmc/tests/test_record.py
+++ b/pytmc/tests/test_record.py
@@ -35,6 +35,31 @@ def test_epics_record_with_linter(dbd_file):
     assert not (linted.errors)
 
 
+def test_input_record_without_write_access():
+    kwargs = {
+        "pvname": "Tst:pv",
+        "record_type": "ai",
+        "direction": "input",
+    }
+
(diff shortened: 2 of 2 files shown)


ZLLentz:
My tiny thoughts on this side are:

- Is it possible to write a useful test for this in `test_record.py`? Well, I can think of some tests to run, but are any of them useful?
- Does this create some backcompat issues while using a newer pytmc to build an old ads-ioc? Is there anywhere here we can or should be checking for version compatibilities?

KaushikMalapati:
> My tiny thoughts on this side are:
> 
> * Is it possible to write a useful test for this in `test_record.py`? Well, I can think of some tests to run, but are any of them useful?
> * Does this create some backcompat issues while using a newer pytmc to build an old ads-ioc? Is there anywhere here we can or should be checking for version compatibilities?

I added some simple test cases. Also, when I edited the jinja template I assumed that direction would always be "input" or "output." Is that true, or can it have any other values like "i," "o," or "io?" 

As far as backward compatibility, I tested that having an ASG field whose value is not specified does not break anything, so there shouldn't be any problems using a newer version of pytmc with an older version of ads-ioc.

ZLLentz:
> Also, when I edited the jinja template I assumed that direction would always be "input" or "output." Is that true, or can it have any other values like "i," "o," or "io?"

Skimming the code, I agree that this should always be "input" or "output" if you go through the full chain but this doesn't seem to be explicitly enforced everywhere... it's worth testing briefly with a real IOC where the user may have used any combinations of these specifiers.

KaushikMalapati:
> > Also, when I edited the jinja template I assumed that direction would always be "input" or "output." Is that true, or can it have any other values like "i," "o," or "io?"
> 
> Skimming the code, I agree that this should always be "input" or "output" if you go through the full chain but this doesn't seem to be explicitly enforced everywhere... it's worth testing briefly with a real IOC where the user may have used any combinations of these specifiers.

https://github.com/pcdshub/pytmc/blob/2c4f4121cd109d2d2cee9bdcc357d792ddca01f7/pytmc/record.py#L498
        record = EPICSRecord(
            pvname,
            self.input_rtyp,
            "input",
            fields=self.field_defaults,
            autosave=self.autosave_defaults.get("input"),
            aliases=aliases,
            archive_settings=self.archive_settings,
            package=self,
        )
...
        record = EPICSRecord(
            self.pvname,
            self.output_rtyp,
            "output",
            fields=self.field_defaults,
            autosave=self.autosave_defaults.get("output"),
            aliases=self.aliases,
            archive_settings=self.archive_settings,
            package=self,
        )
Looks like it will always be "input" or "output"

ZLLentz:
I do not know why you're failing the test on python 3.12 but I do know that it probably has nothing to do with this PR

ZLLentz (approved):
The pytmc diff here is simple and effective when paired with the ads-ioc PR

ZLLentz:
I'm happy to merge this and get a new pytmc tag out if we're satisfied here

PR #331, opened 2024-11-20
FIX: workaround for files that end up in strange subfolders

ZLLentz:
This is to address https://jira.slac.stanford.edu/browse/ECS-6685

This works offline and doesn't seem to break the test suite
I'd like to deploy this as a tag for use in the CI builds

### DIFF
--- a/pytmc/parser.py
+++ b/pytmc/parser.py
@@ -147,7 +147,7 @@ def element_to_class_name(
     return tag, TwincatItem
 
 
-def _determine_path(base_path, name, class_hint):
+def _determine_path(base_path, name, class_hint) -> pathlib.Path:
     """
     Determine the path to load child XTI items from, given a base path and the
     class load path hint.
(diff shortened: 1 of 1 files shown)


ZLLentz:
conda CI is failing during the build step, I guess I can investigate that in a separate PR

ZLLentz:
@tangkong does this look dangerous to you? It didn't break any of the tests and it seemed to help with the linked issue. @yannS2016 and @NSLentz were starting to deploy some PLC library updates and having a fix for their issue deployed would help a lot. I'm tentatively planning to merge and introduce a bugfix tag.

ZLLentz:
Side-note: it seems like we don't have a great way to run this in CI in a dev sense like we do with some of the other CI builds.

tangkong:
If this helps remedy the problem I'm ok with them using it, but I do worry about how this might interact with files with identical names that are differentiated by the folder structure.  I'm sure we have some confidence that this doesn't happen, but I don't immediately know enough about the twincat file creation to understand why. 

Is it not possible to deploy this as a dev hotfix in Yann/NLentz's home areas for use until we fully understand why the folder structure changes?  This is one step away from "search the entire folder recursively for X filename", which worries me.

If twincat-math is the only project that's doing this, that makes me suspect that project specifically.  Rather than thinking that `pytmc` needs to be more permissive

tangkong:
> Side-note: it seems like we don't have a great way to run this in CI in a dev sense like we do with some of the other CI builds.

As in we don't have a `pytest` equivalent to run for PLC projects?  I do think that'd be nice to have

tangkong (approved):
I'll drop an approval here if we deem the best course of action is to merge>tag>deploy, but I still think that we should just reference a local copy of pytmc if we can

ZLLentz:
I have no idea what happened here but this PR merged in a malformed way. I need to make a follow-up.

ZLLentz:
I must have tested a slightly different version than what got included here.

PR #332, opened 2024-12-01
Support for arrays of strings

### LINKED ISSUE #327: Add support for arrays of strings
Currently, if you place a pytmc pragma on an array of strings, it will error out with:

```
ERROR:pytmc.bin.template:Failed to create EPICS records
Traceback (most recent call last):
File "c:\repos\pytmc\pytmc\bin\template.py", line 374, in get_plc_record_packages
packages, exceptions = db_process(
File "c:\repos\pytmc\pytmc\bin\db.py", line 154, in process
record_names = [
File "c:\repos\pytmc\pytmc\bin\db.py", line 158, in
for single_record in record_package.records
File "c:\repos\pytmc\pytmc\record.py", line 619, in records
records = [self.generate_input_record()]
File "c:\repos\pytmc\pytmc\record.py", line 955, in generate_input_record
record = super().generate_input_record()
File "c:\repos\pytmc\pytmc\record.py", line 511, in generate_input_record
record.fields["DTYP"] = self.dtyp
File "c:\repos\pytmc\pytmc\record.py", line 952, in dtyp
return data_types[self.chain.data_type.name]
KeyError: 'STRING'
```

This is because it is entering the `WaveformRecordPackage` and trying to create a waveforms of strings, which is intentionally not supported.

However, it would be better if these were handled more like arrays of structs and created a separate string PV for each array element.

See the option assembly code at https://github.com/pcdshub/pytmc/blob/2c4f4121cd109d2d2cee9bdcc357d792ddca01f7/pytmc/pragmas.py#L322-L327 for how this is handled for structured arrays.

KaushikMalapati:
Meant to close #327. I made two small changes for this to work

- In from chain, I moved the data_type.is_string in front of the data_type.is_array or chain.last.array_info so that arrays will not use WaveformRecordPackage.
- I handle string arrays as a complex type instead of a non-complex type so that a pv is made for each array index

### DIFF
--- a/pytmc/pragmas.py
+++ b/pytmc/pragmas.py
@@ -320,7 +320,7 @@ def _expand_configurations_from_chain(
             return []
 
         if item.array_info and (
-            item.data_type.is_complex_type or item.data_type.is_enum
+            item.data_type.is_complex_type or item.data_type.is_enum or item.data_type.is_string
         ):
             options = get_all_options(subitems, handle_array_complex, pragmas)
         else:
--- a/pytmc/record.py
+++ b/pytmc/record.py
@@ -298,10 +298,10 @@ class RecordPackage:
             spec = RecordPackage
         elif data_type.is_enum:
             spec = EnumRecordPackage
-        elif data_type.is_array or chain.last.array_info:
-            spec = WaveformRecordPackage
         elif data_type.is_string:
             spec = StringRecordPackage
+        elif data_type.is_array or chain.last.array_info:
+            spec = WaveformRecordPackage
         else:
             try:
                 spec = DATA_TYPES[data_type.name]
--- a/pytmc/tests/test_xml_collector.py
+++ b/pytmc/tests/test_xml_collector.py
@@ -107,6 +107,7 @@ def chain():
         ("LREAL", False, FloatRecordPackage),
         ("LREAL", True, WaveformRecordPackage),
         ("STRING", False, StringRecordPackage),
+        ("STRING", True, StringRecordPackage),
     ],
 )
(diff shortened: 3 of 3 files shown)


tangkong:
Conda test failures seem unrelated, we need to add setuptools to the recipe since it's no longer implicit 
https://github.com/pcdshub/pytmc/actions/runs/12101972735/job/33797497583?pr=332#step:12:216

tangkong (commented):
This looks reasonable to me but I will defer to Zach here.

tangkong:
Yes, but I would prefer we make a new PR for this as to be very clear when we modify build recipes.  See our other python repos for example conda recipes (we added setuptools_scm gradually as we noticed these errors iirc)
https://github.com/pcdshub/pcdsdevices/blob/6a4e36fbeed085ee3f34ee327a41fcc9e9e7eae4/conda-recipe/meta.yaml#L21

ZLLentz:
The changes here make sense in principle but this sort of thing really needs a unit test.

For example, what about adding this new case to the parameterization in https://github.com/pcdshub/pytmc/blob/c30a7c9adc24e1c77150248f330d80c67d845a82/pytmc/tests/test_xml_collector.py#L94-L111 ?
(just as one example)

I wonder if there are other places in the test suite that make sense to extend to check the new feature. Perhaps it's worth adding something later in this test_xml_collector near where we check other kinds of "complex" arrays?

I'm not exactly sure what this looks like in practice but the lack of testing is the thing that stands out most here.

ZLLentz:
I like the tests you added a lot!

I'm going to click "update branch" and if it passes we should merge this!

ZLLentz (approved):
This is a good expansion of our IOC gen capabilities.

PR #333, opened 2024-12-05
Adding setuptools_scm to conda recipe

KaushikMalapati:
Noticed that conda tests were failing because setuptools wasn't implicitly present in #332.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -15,7 +15,7 @@ build:
   script: {{ PYTHON }} -m pip install . -vv
 
 requirements:
-  build
+  build:
   - setuptools_scm
   host:
   - python >=3.9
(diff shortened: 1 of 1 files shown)


tangkong:
Don't worry about re-running actions, you should feel free to consume as many github runner minutes as you please 😄 

The nuances of the conda recipies are sometimes lost on me, my best explanation is that setuptools was implicitly installed by conda (either via another package or bundled with conda itself), and this changed at some point.  Reading through documentation one would expect that the `build` section would also work, admittedly we simply followed the guidance on conda-forge https://github.com/conda-forge/pcdsdevices-feedstock/pull/63#issuecomment-2311424237  

Apparently this is a pre-requisite for packages with pip dependencies

tangkong (approved):

PR #335, opened 2025-02-12
FIX: remove wayward duplicate line from pre-fix

ZLLentz:
Follow-up to #331 

I made an error there that caused my fix to be completely skipped.

I've also updated the release notes docs in this PR.

I'm doing one more offline test before I even consider hitting merge.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -2,6 +2,21 @@
  Release History
 =================
 
+v2.18.2 (2025-02-12)
+====================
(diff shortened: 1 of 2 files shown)


tangkong (approved):
Ah, I'm sorry I should have caught that

ZLLentz:
Confirmed that `pytmc summary` completes for https://github.com/pcdshub/lcls-twincat-math/pull/5 with this activated and fails without

PR #337, opened 2025-02-20
FIX: also allow BYTE arrays to be waveform records

ZLLentz:
Prospective update as a possible way for us to get long long long long strings from a PLC


ZLLentz:
This did work in testing when I made it- since it's so simple and the main consequence is enabling an optional feature, I'm inclined to resolve the conflicts and get it in

ZLLentz:
Yeah, this came in as part of another PR. Rejoice!

PR #339, opened 2025-10-06
Added Support for non EtherCATMC type motors

yannS2016:
Added Support for non-EtherCATMC-type motors

### DIFF
--- a/pytmc/bin/template.py
+++ b/pytmc/bin/template.py
@@ -26,7 +26,8 @@ The following helpers are available in the environment::
     get_library_versions
     get_links
     get_linter_results
-    get_motors
+    get_legacy_motors
+    get_stream_motors
     get_nc
     get_plc_by_name
     get_pou_call_blocks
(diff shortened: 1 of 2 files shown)


pytmc/parser.py
```diff
@@ -1661,6 +1661,111 @@ def nc_axis(self):
 Symbol_DUT_MotionStage = Symbol_ST_MotionStage
 
 
+
+class Symbol_FB_MotionStage(Symbol):
+    """[TMC] Customized Symbol representing only ST_MotionStage.
```
ZLLentz, line 1666:
```suggestion
    """[TMC] Customized Symbol representing only FB_MotionStage.
```

pytmc/parser.py
```diff
@@ -1661,6 +1661,111 @@ def nc_axis(self):
...
+        except Exception as ex:
+            repr_info["nc_axis"] = repr(ex)
+        return repr_info
+
+    def get_axis_link(self):
+        """
+        Returns the first axis-link value from a pytmc pragma (e.g. 'GVL_Axes.Axes[2]')
+        Raises RuntimeError if not found.
+        """
+        from .pragmas import get_pragma
```
ZLLentz, line 1688: Suggestion: imports always at the top of the file is the convention

pytmc/parser.py
```diff
@@ -1661,6 +1661,111 @@ def nc_axis(self):
...
+        for pragma_entry in get_pragma(self, name='pytmc'):
```
ZLLentz, line 1689: Question: how does this function work? I don't understand how a pytmc pragma could ever have axis linking information, nor how we can ensure that this is how the link is always configured.

pytmc/parser.py
```diff
@@ -1661,6 +1661,111 @@ def nc_axis(self):
...
+
+    @property
+    def nc_to_plc_link(self):
+        """
+        Returns the Link whose 'a[1]' contains the axis-link (and '.NcToPlc').
+        Raises RuntimeError for mismatches.
+        """
+        plc = getattr(self, "plc", None)
+        if plc is None:
+            return None
```
ZLLentz, line 1722: Question: why does this sometimes return None and sometimes raise? Shouldn't it always raise as per the docstring?

pytmc/parser.py
ZLLentz, line 1: Suggestion: there's a lot of code added in this PR but no unit tests. We need to add some unit tests to cover the new features.

ZLLentz (changes requested):
I think we can't merge this unless it passes the CI and includes testing for the added features.

yannS2016:
> Please go through the tests and address the failures. It looks like your modifications have broken the expected behaviour of the cli entrypoint.
> 
> There are several pre-commit problems to clean up as well. If you install pre-commit locally, you can catch these mistakes before committing and pushing

Thought I mark this as a draft! Was tracking the issue with CLI!

yannS2016:
Thanks for the initial! Review, I meant to mark this as a draft to address the CLI issue! Other cleanup will follow!

yannS2016:
> Please go through the tests and address the failures. It looks like your modifications have broken the expected behaviour of the cli entrypoint.
> 
> There are several pre-commit problems to clean up as well. If you install pre-commit locally, you can catch these mistakes before committing and pushing

Sure, this was meant to be a draft, not for review.

pytmc/bin/template.py
```diff
@@ -312,15 +313,68 @@ def get_symbols_by_type(plc: parser.Plc) -> dict[str, list[parser.Symbol]]:
...
-    return [
-        stage
-        for stage in symbols.get("Symbol_ST_MotionStage", [])
-        if not stage.is_pointer and pragmas.has_pragma(stage)
-    ]
+    st_motors = ensure_list(symbols.get("Symbol_ST_MotionStage", []))
+    legacy = []
+    for stage in st_motors:
+        if (
+            not getattr(stage, "is_pointer", False)
```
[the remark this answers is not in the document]
yannS2016: Agreed, @tangkong, in this instance, we do not want to include motors that do not have the is_pointer attribute (set or not). Thus, we can directly test the variable since it's local to the instance.

yannS2016:
I think most comments and suggestions here have been addressed for this second iteration. I'm pushing a couple of new commits, including @ZLLentz suggestion about moving 'has_pragma and get_pragma' to the parser script. The changes are light and do not break tests.

ZLLentz:
I think it looks good minus the duplicated submodule

Two things I'm checking before I approve:

- Verifying that the existing tests are sufficient
- Verifying that an old IOC still builds OK with the newer pytmc from this PR

ZLLentz:
Confirmed that an old IOC builds exactly the same!
Looking at the testing next

yannS2016:
> I'll approve after we figure out what's going on with the `.gitmodules`
> 
> This is unit testing just as much after this PR as before so I can't really fault the PR 🤷
> 
> I'm really upset at past me for not insisting on more thorough unit testing of pytmc template- it's mostly just a couple surface-level tests in https://github.com/pcdshub/pytmc/blob/master/pytmc/tests/test_commandline.py

Seems i had a duplicate of the submodule. should be fine now.

